PR #1, opened 2022-06-10
Changed CMake minimal version to 3.16

iTrooz:
The minimal CMake version is currently 3.21, but the project can be compiled using older versions, such as 3.16

This should (kind of) fix this issue : https://github.com/WerWolv/ImHex/issues/506


WerWolv:
Thanks! CLion automatically sets it to the current version and I forgot to lower it

PR #4, opened 2022-07-12
Pointer fixes & feature

PredatorCZ:
A few updates.

- Fixed crash when setting color override attribute
- Pointer itself is catchable via getPattern (can set bg color in editor)
- Pointer can point to array
 
So instead of

```rust
struct Array {
  u32 data[10];
};

Array *arr : u32 @0;
```

can simply be

```rust
u32 *data[10] : u32 @0;
```

Can add unit test later :)


WerWolv (approved):

WerWolv:
That's really really cool. Thanks a lot!

PR #5, opened 2022-07-14
pattern: pointer value can be signed

PredatorCZ:
Hi, I added support for signed pointer values, they are used for relative pointers.

### DIFF
--- a/lib/include/pl/ast/ast_node_attribute.hpp
+++ b/lib/include/pl/ast/ast_node_attribute.hpp
@@ -157,14 +157,14 @@ namespace pl {
                 LogConsole::abortEvaluation("pointer base function needs exactly one parameter", node);
 
             if (auto pointerPattern = dynamic_cast<PatternPointer *>(pattern)) {
-                u128 pointerValue = pointerPattern->getPointedAtAddress();
+                i128 pointerValue = pointerPattern->getPointedAtAddress();
 
                 auto result = function.func(evaluator, { pointerValue });
 
                 if (!result.has_value())
                     LogConsole::abortEvaluation("pointer base function did not return a value", node);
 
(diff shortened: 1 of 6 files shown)


lib/include/pl/ast/ast_node_attribute.hpp
```diff
@@ -157,14 +157,14 @@ namespace pl {
                 LogConsole::abortEvaluation("pointer base function needs exactly one parameter", node);
 
             if (auto pointerPattern = dynamic_cast<PatternPointer *>(pattern)) {
-                u128 pointerValue = pointerPattern->getPointedAtAddress();
```
WerWolv, line 160: This will break if you're using a u128 pointer size since it might overflow.

Since it's very unlikely that anybody will ever use u128s as the size though, we might want to just restrict the max size of pointers to an u96/s96

lib/include/pl/patterns/pattern_pointer.hpp
```diff
@@ -67,18 +69,29 @@ namespace pl {
...
+        [[nodiscard]] i128 getPointedAtAddress() const {
             return this->m_pointedAtAddress;
         }
 
         [[nodiscard]] const std::shared_ptr<Pattern> &getPointedAtPattern() {
             return this->m_pointedAt;
         }
 
+        [[nodiscard]] bool isSigned() const {
+            return this->isSigned();
```
WerWolv, line 92: This is an infinite recursion loop. You probably meant to use `this->m_signed` here

WerWolv (changes requested):

lib/include/pl/ast/ast_node_attribute.hpp
PredatorCZ: I've limited pointer size to max 64 bits, if you don't mind.

WerWolv (approved):

WerWolv:
Thank you!

PR #7, opened 2022-07-16
pattern: inlined array shows varname

PredatorCZ:
Inlined array should now show as `<varname>[<index>]` instead of `[<index>]`

### DIFF
--- a/lib/include/pl/patterns/pattern_array_static.hpp
+++ b/lib/include/pl/patterns/pattern_array_static.hpp
@@ -23,9 +23,18 @@ namespace pl {
             auto entry = std::shared_ptr(this->m_template->clone());
             for (u64 index = 0; index < std::min<u64>(end, this->m_entryCount); index++) {
                 entry->clearFormatCache();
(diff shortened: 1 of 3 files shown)


lib/include/pl/patterns/pattern_array_static.hpp
```diff
@@ -23,12 +23,21 @@ namespace pl {
...
                 fn(index, *entry);
             }
         }
 
+        void setInlined(bool isInlined) override {
+            Inlinable::setInlined(isInlined);
+            if (isInlined) {
+                this->m_template->setVariableName(this->getVariableName());
+            } else {
+                this->m_template->setVariableName("");
```
PredatorCZ, line 37: I wonder, if it would be better to store format in varname instead of string cat at draw time.

WerWolv:
I'm so sorry, I let this sit around for WAY WAY too long
Things have changed quite a bit since and I'd probably implement this as part of the ImHex pattern drawer, maybe as an option. Thanks for the PR though

PR #14, opened 2022-10-20
Use std::make_shared when creating std::shared_ptr

tnixeu:
I was looking through the code of ImHex and noticed in this submodule this mixture of smart pointers.

In a lot of places I think `std::unique_ptr` is used because it is unclear if it it will be assigned to a `std::unique_ptr` or  `std::shared_ptr`. However, in this case the assignment is always towards a `std::shared_ptr`. 

Is the code in PatternLanguage important for the performance of ImHex?


WerWolv:
Thanks! I think this was a leftover from a refactor I did a while ago.

> Is the code in PatternLanguage important for the performance of ImHex?
Yeah the Pattern Language should run as fast as possible so big files can be parsed quickly and efficiently. The code you modified only applies to creating variables using the `auto` type though so it's not _super_ critical

PR #15, opened 2022-10-26
Use std::make_* instead of constructors

tnixeu:
I wanted to see, how often a `std::unique_ptr` is assigned to a `std::shared_ptr`, which I now know that it happens only in a few places. 

This PR changes the `create` function in order to create the smart pointers directly instead of taking a pointer and also splits the function into two version. `create` still returns a `std::unique_ptr`, whereas `createShared` now returns a `std::shared_ptr`.

This PR also changes the `moveToVector` in order to use a fold expression instead of a recursion. It is such a small changed, that I did not want to open another PR for it.

### DIFF
--- a/lib/include/pl/core/parser.hpp
+++ b/lib/include/pl/core/parser.hpp
@@ -48,10 +48,18 @@ namespace pl::core {
             return this->m_curr[index].column;
         }
 
-        template<typename T>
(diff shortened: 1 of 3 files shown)


WerWolv:
Thank you!

PR #18, opened 2022-11-04
Allow unlimited limits (== 0)

ctrlcctrlv:


WerWolv:
Thank you for the PR!
Could you add some more information about how this change works and can be used?

lib/source/pl/lib/std/pragmas.cpp
```diff
@@ -3,8 +3,24 @@
 #include <pl/core/validator.hpp>
 #include <pl/core/evaluator.hpp>
 
-namespace pl::lib::libstd {
+const long unsigned int LIMITMAX = std::numeric_limits<long unsigned int>::max() - 1;
```
WerWolv, line 6: This should probably be `constexpr static auto MaxLimit = std::numeric_limits<u64>::max();`

lib/source/pl/lib/std/pragmas.cpp
```diff
@@ -3,8 +3,24 @@
...
+
+bool parse_limit(const std::string &value, long unsigned int *limit) {
```
WerWolv, line 8: Camel case to match the rest of the code.
Also use `u64` and so on for type names.
I'd probably instead return a `std::optional<u64>` here where an empty optional signalizes that an error occurred instead of doing it through an out parameter

lib/source/pl/lib/std/pragmas.cpp
```diff
@@ -3,8 +3,24 @@
...
+    char* s = (char*)value.c_str();
```
WerWolv, line 9: This is undefined behaviour. You can't cast away const

lib/source/pl/lib/std/pragmas.cpp
```diff
@@ -3,8 +3,24 @@
...
+    char** endptr = &s;
+    auto tlimit = strtol(s, endptr, 0);
```
WerWolv, line 11: `std::stoll` might be the better alternative so you don't need to do this stuff right above

lib/source/pl/lib/std/pragmas.cpp
```diff
@@ -21,9 +37,10 @@ namespace pl::lib::libstd {
         });
 
         runtime.addPragma("eval_depth", [](pl::PatternLanguage &runtime, const std::string &value) {
-            auto limit = strtol(value.c_str(), nullptr, 0);
+            long unsigned int limit = 0;
+            bool parsed = parse_limit(value, &limit);
 
-            if (limit <= 0)
+            if (!parsed || limit == LIMITMAX)
```
WerWolv, line 43: When you use an optional for the function above, it might look something like 

```cpp
if (auto limit = parseLimit(value); !limit.has_value())
    return false;
else
    runtime.getInternals().evaluator->setEvaluationDepth(limit.value());
```

WerWolv (changes requested):

ctrlcctrlv:
sorry for all the issues. i don't write much c++.

WerWolv:
No worries at all! I just wanted to give you some feedback. Thanks a lot for the effort :)

ctrlcctrlv:
>  Could you add some more information about how this change works and can be used?

This change was written directly to allow this:

https://github.com/WerWolv/ImHex-Patterns/blob/3065540aab760e43b84d1a0264a78cd23dc83439/patterns/png.hexpat#L3

```c++
#pragma pattern_limit 0
```

Without this change, it is returned in ImHex that:

Screenshot from 2022-11-05 05-57-04

(I don't know what's up with the wrong line being highlighted, probably an existing bug.)

gmestanley:
The wrong line being highlighted is indeed a bug, but that's due to the third-party editor [ImGuiColorTextEdit](https://github.com/BalazsJako/ImGuiColorTextEdit) being used. As for what's being highlighted, I think a limit of 0 for patterns not being allowed is intentional due to the reasons listed [here](https://github.com/WerWolv/ImHex-Patterns/pull/53#discussion_r1013795167). Generally pattern coders just set a pattern limit sufficiently higher than the default (I use numbers that set bits like 0x40000) if there's a lot of patterns.

ctrlcctrlv:
Giving people the option is needed and useful. I've used ImHex to edit 70MB APNG files.

WerWolv:
I'm okay with adding support for it to the pattern language but I think it should not be added to "official" patterns directly. It's up to the user to increase that limit if it's necessary for the file they're trying to load. Otherwise ImHex will just crash right away if the user loads a huge pattern on an old machine

WerWolv:
Thank you!

PR #20, opened 2023-01-09
Conditionally set CXX_FLAGS for gcc and clang

supervacuus:
Hi, thank you for your great editor. I tried to build ImHex on clang, but `PatternLanguage` fails because clang currently doesn't support `stringop-overflow`. Also, both flags won't work on MSVC either.


supervacuus:
I am confident that the failing test is unrelated to my change. But if there is a quick fix that I can do, let me know.

WerWolv:
Thanks for the fixes! I don't really support MSVC anyways so that's not a concern.

Yeah the build fails because I seem to have configured the CI incorrectly somewhere, it's definitely not your fault.

PR #22, opened 2023-02-16
`Match` keyword for rust like match

Nowilltolife:
## Added features:
Struct match feature:
```rust
struct a {
	u8 badge;
        u8 type;
	match(badge, type) {
                (0x10, 0x20) : {
                   u8 size; 
                }
                (0x10, _) : {
                   u16 completeSize;
                 }
	}
};
```

Function match feature:
```rust
#include <std/io.pat>

fn main() {
    u8 type @ 0x0;
    match(type) {
        (0) : std::print("Type is 0");
        (1) : std::print("Type is 1");
        (_) : std::print("Unknown type");
    }
};
```

Syntax:
`match (expression**) { ((literal**)  : statement?|{statement**} }`

### DIFF
--- a/lib/include/pl/core/ast/ast_node_match_statement.hpp
+++ b/lib/include/pl/core/ast/ast_node_match_statement.hpp
@@ -26,17 +26,25 @@ namespace pl::core::ast {
             this->condition = std::move(other.condition);
             this->body = std::move(other.body);
         }
+
+        MatchCase &operator=(const MatchCase &other) {
+            this->condition = other.condition->clone();
+            for (auto &statement : other.body)
+                this->body.push_back(statement->clone());
+            return *this;
+        }
     };
 
     class ASTNodeMatchStatement : public ASTNode {
 
     public:
-        explicit ASTNodeMatchStatement(std::vector<MatchCase> cases)
-            : ASTNode(), m_cases(std::move(cases)) { }
+        explicit ASTNodeMatchStatement(std::vector<MatchCase> cases, std::optional<MatchCase> defaultCase)
+            : ASTNode(), m_cases(std::move(cases)), m_defaultCase(std::move(defaultCase)) { }
 
         ASTNodeMatchStatement(const ASTNodeMatchStatement &other) : ASTNode(other) {
             for (auto &matchCase : other.m_cases)
                 this->m_cases.push_back(matchCase);
+            other.m_defaultCase ? this->m_defaultCase.emplace(*other.m_defaultCase) : this->m_defaultCase = std::nullopt;
         }
 
         [[nodiscard]] std::unique_ptr<ASTNode> clone() const override {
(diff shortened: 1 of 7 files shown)


lib/include/pl/core/ast/ast_node_match_statement.hpp
```diff
@@ -0,0 +1,156 @@
...
+    class ASTNodeMatchStatement : public ASTNode {
+
+    public:
+        explicit ASTNodeMatchStatement(std::vector<MatchCase> cases, std::optional<MatchCase> defaultCase)
+            : ASTNode(), m_cases(std::move(cases)), m_defaultCase(std::move(defaultCase)) { }
+
+        ASTNodeMatchStatement(const ASTNodeMatchStatement &other) : ASTNode(other) {
+            for (auto &matchCase : other.m_cases)
+                this->m_cases.push_back(matchCase);
+            other.m_defaultCase ? this->m_defaultCase.emplace(*other.m_defaultCase) : this->m_defaultCase = std::nullopt;
```
WerWolv, line 47: This should be a plain old if-else statement instead of a ternary

lib/include/pl/core/ast/ast_node_match_statement.hpp
```diff
@@ -0,0 +1,156 @@
...
+            return std::unique_ptr<ASTNode>(new ASTNodeMatchStatement(*this));
+        }
+
+        [[nodiscard]] std::vector<std::shared_ptr<ptrn::Pattern>> createPatterns(Evaluator *evaluator) const override {
+            evaluator->updateRuntime(this);
+
+            auto &scope = *evaluator->getScope(0).scope;
+            auto body  = getCaseBody(evaluator);
+
+            if(!body) return {};
```
WerWolv, line 60: There's a few occurrences of this. I always have a space between `if`, `while`, `for`, `switch`, etc keywords and the `(` of the head

lib/include/pl/core/ast/ast_node_match_statement.hpp
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+         FunctionResult execute(Evaluator *evaluator) const override {
+            evaluator->updateRuntime(this);
+
+            auto body = getCaseBody(evaluator);
+
+            if(!body) return std::nullopt;
+
+            auto variables     = *evaluator->getScope(0).scope;
```
WerWolv, line 82: You're calling `evaluator->getScope(0)` quite often here. Consider getting a reference to it at the start and then just use that

lib/include/pl/core/ast/ast_node_match_statement.hpp
```diff
@@ -0,0 +1,156 @@
...
+            return std::visit(hlp::overloaded {
+                [](const std::string &value) -> bool { return !value.empty(); },
+                [this](ptrn::Pattern *const &pattern) -> bool { err::E0004.throwError(fmt::format("Cannot cast value of type '{}' to type 'bool'.", pattern->getTypeName()), {}, this); },
+                [](auto &&value) -> bool { return value != 0; }
+            }, literal->getValue());
+        }
+
+        [[nodiscard]] const std::vector<std::unique_ptr<ASTNode>>* getCaseBody(Evaluator *evaluator) const {
+            std::optional<size_t> matchedBody;
+            for(size_t i = 0; i < m_cases.size(); i++) {
```
WerWolv, line 137: Use `this->m_cases` to access members

lib/include/pl/core/token.hpp
```diff
@@ -48,6 +48,8 @@ namespace pl::core {
             Parent,
             This,
             While,
+            Match,
+            Default,
```
WerWolv, line 52: The default keyword isn't used now and can be removed again

lib/source/pl/core/parser.cpp
```diff
@@ -724,6 +768,95 @@ namespace pl::core {
         return create<ast::ASTNodeConditionalStatement>(std::move(condition), std::move(trueBody), std::move(falseBody));
     }
 
+    std::pair<std::unique_ptr<ast::ASTNode>, bool> Parser::parseCaseParameters(std::vector<std::unique_ptr<ast::ASTNode>> &condition) {
+        std::vector<std::unique_ptr<ast::ASTNode>> compiledConditions;
+
+        size_t i = 0;
```
WerWolv, line 774: This should probably have a better name such as `caseIndex` or `caseNumber`

lib/source/pl/core/parser.cpp
```diff
@@ -724,6 +768,95 @@ namespace pl::core {
...
+            err::P0002.throwError("No parameters found in case statement.", {}, 1);
+        }
+
+        if(i != condition.size()) {
+            err::P0002.throwError("Size of case parameters smaller than size of match condition.", {}, 1);
+        }
+
+        // make multi expressions into single expression
+        auto cond = std::move(compiledConditions[0]);
+        for(size_t j = 1; j < compiledConditions.size(); j++) {
```
WerWolv, line 811: This might also benefit from a better name

WerWolv (changes requested):

WerWolv (approved):

PR #27, opened 2023-02-18
Added math::accumulate

Nowilltolife:
## What's new
New math::accumulate function allowing to add, multiply, min, max or modulo over a range, mostly targeted to checksum
functions.
Definition:
```rust
enum Operation : u8 {
   Add,
   Mul,
   Min,
   Max,
   Mod
};

std::math::accumulate(u128 from. u128 to, u128 size, Operation operation = Operation::ADD,  Endian endian = Endian::Native) -> u128
```

## Example
```rust
#include <std/math.pat>
#include <std/io.pat>

fn main() {
   u32 checksum @ 0x00;
   u32 sum = std::math::accumulate(0x00, std::mem::size(), sizeof(u8));
   if(sum != checksum) {
        std::print("Checksum incorrect");
   }
}
```

### DIFF
--- a/lib/source/pl/lib/std/math.cpp
+++ b/lib/source/pl/lib/std/math.cpp
@@ -1,3 +1,5 @@
+#include <numeric>
+
 #include <pl.hpp>
 
 #include <pl/core/token.hpp>
@@ -5,20 +7,16 @@
 #include <pl/core/evaluator.hpp>
 #include <pl/patterns/pattern.hpp>
 #include <pl/helpers/buffer.hpp>
-#include <numeric>
-#include "pl/lib/std/types.hpp"
-
-#define BUFFER_THRESHOLD 0x3000
-#define CHUNKS 20
+#include <pl/lib/std/types.hpp>
 
 namespace pl::lib::libstd::math {
 
-    enum AccumulationOperation {
-        ADD,
-        MULTIPLY,
-        MODULO,
-        MIN,
-        MAX
+    enum class AccumulationOperation {
+        Add,
+        Multiply,
+        Modulo,
+        Min,
+        Max
     };
 
     void registerFunctions(pl::PatternLanguage &runtime) {
@@ -155,7 +153,7 @@ namespace pl::lib::libstd::math {
                 auto start = params[0].toUnsigned();
                 auto end = params[1].toUnsigned();
                 auto size = params[2].toUnsigned();
-                AccumulationOperation op = ADD;
+                AccumulationOperation op = Add;
                 if(params.size() > 3) {
                     op = static_cast<AccumulationOperation>(params[3].toUnsigned());
                 }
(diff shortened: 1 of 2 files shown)


lib/source/pl/lib/std/math.cpp
```diff
@@ -4,9 +4,23 @@
 #include <pl/core/log_console.hpp>
 #include <pl/core/evaluator.hpp>
 #include <pl/patterns/pattern.hpp>
+#include <pl/helpers/buffer.hpp>
+#include <numeric>
```
WerWolv, line 8: Please use `<>` for all includes and separate regular includes from stdlib includes

lib/source/pl/lib/std/math.cpp
```diff
@@ -4,9 +4,23 @@
...
+#include "pl/lib/std/types.hpp"
+
+#define BUFFER_THRESHOLD 0x3000
```
WerWolv, line 11: Defines should instead be `constexpr static` variables

lib/source/pl/lib/std/math.cpp
```diff
@@ -4,9 +4,23 @@
...
+#define CHUNKS 20
 
 namespace pl::lib::libstd::math {
 
+    enum AccumulationOperation {
```
WerWolv, line 16: Use `enum class` and `PascalCase` for all enum (but also other) constants

lib/source/pl/lib/std/math.cpp
```diff
@@ -136,6 +149,45 @@ namespace pl::lib::libstd::math {
...
+                    op = static_cast<AccumulationOperation>(params[3].toUnsigned());
+                }
+                types::Endian endian = 0;
+                if(params.size() > 4) {
+                    endian = static_cast<types::Endian>(params[4].toUnsigned());
+                }
+
+                u128 result = 0;
+                u128 endAddr = end / size;
+                using namespace hlp::bf;
```
WerWolv, line 169: Probably wouldn't use a `using namespace` here since it's just used for one single type

lib/source/pl/lib/std/math.cpp
```diff
@@ -136,6 +149,45 @@ namespace pl::lib::libstd::math {
...
+                auto reader = BufferedReader(ctx);
+                reader.seek(start);
+                reader.setEndAddress(end);
+                for(u128 addr = start; addr < endAddr; ++addr) {
+                    auto bytes = reader.read(addr, size);
+                    // copy to u128
+                    u128 value = 0;
+                    std::memcpy(&value, bytes.data(), size);
```
WerWolv, line 177: Make sure to limit `size` to be 16 at max, otherwise you'll get a potential buffer overflow here

WerWolv (changes requested):

PR #28, opened 2023-02-25
Allow bitfields to use conditionals correctly, and fix big-endian bitfield oddities.

Zaggy1024:
This changes big-endian (`left_to_right`) bitfields to be read in big-endian order, which allows variables to be referenced in the logical order that they are declared. In the process, this also fixes an issue where bitfields would skip the first bits in a non-byte-sized big-endian bitfields to align it to the rightmost byte.

Bitfields' fields are now tracked in descendant AST nodes, meaning that bitfield offsets for conditional fields will no longer be zero. This should allow conditional reads in bitfields to work as one would expect.

### DIFF
--- a/lib/include/pl/core/ast/ast_node_bitfield.hpp
+++ b/lib/include/pl/core/ast/ast_node_bitfield.hpp
@@ -2,6 +2,7 @@
 
 #include <pl/core/ast/ast_node.hpp>
 #include <pl/core/ast/ast_node_attribute.hpp>
+#include <pl/core/ast/ast_node_bitfield_field.hpp>
 
 #include <pl/patterns/pattern_bitfield.hpp>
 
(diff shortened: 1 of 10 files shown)


Zaggy1024:
Converting to a draft since I noticed a regression in the `right_to_left` behavior.

Zaggy1024:
Should be good to review, I had to rework the first commit quite a bit to get things all working correctly, so here's the first commit summarizing the changes that were needed:

> Correct the byte order of bitfields, don't reverse big-endian fields.
> 
> This fixes two issues:
> - Bitfields would be passed the data with all bytes reversed vs what the reads would require, which caused multi-byte bitfields to read incorrectly, especially if they are not exactly byte-sized.
> - Big-endian bitfields would have their AST node children reversed to read in that order, but this would cause non-byte-sized bitfields to read aligned to the right byte instead of the left, and conditional statements would not work at all.

Zaggy1024:
Just rebased this onto latest.

WerWolv:
Thanks! I reviewed it before and it looks good.
The one question I have though, what's the reason for that callback you added? Can't you just execute that code right after calling createPatterns on the bitfield field ast node inside the bitfield ast node?

WerWolv:
Oh the bitfield unit test also seem to be failing right now. It probably also needs to be adjusted to reflect the new behaviour

Zaggy1024:
> Thanks! I reviewed it before and it looks good. The one question I have though, what's the reason for that callback you added? Can't you just execute that code right after calling createPatterns on the bitfield field ast node inside the bitfield ast node?

It's been a bit, but I believe at the very least nested conditionals wouldn't work right without this, since the bit offsets wouldn't be valid until after the conditional branch's patterns were already emitted, so any references to bitfield fields wouldn't evaluate correctly.

Now that I'm thinking about it, though, to add nested bitfields it might be necessary to store a current bit offset in Evaluator instead of a callback. I'll probably have to revisit this, I might come up with a more elegant solution.

WerWolv:
Oh yeah that makes sense, I didn't think about conditionals. I'm okay with merging this PR once the unit tests have been fixed though and you can improve the current system with future PRs

Zaggy1024:
Looks like TestPatternBitfields is failing with these changes, though, I'm guessing with limited understanding of the testing that it's because it uses `be` on the bitfield variable declaration. I think that needs to be `le` since that would match the previous behavior, so I'll try pushing that as a fix and check when I wake up.

Zaggy1024:
Got the tests building locally and fixed up the rest of the issues with that test, should be good now.

WerWolv:
Thank you!

PR #31, opened 2023-03-15
Allow inheriting of template types, prevent some undefined behavior

Zaggy1024:
This splits off parsing of custom type declarations from normal type declarations to allow template parameters to be parsed using that function. Using that function allows more verification of the types that are inherited from and to read template parameter values. For example, the following code will now work:

```c++
struct Base<T> {
    T parent_var;
};

struct Test<T> : Base<T> {
    T child_var;
};

Test<u32> test @ 0;
```

In the process, several issues should be fixed:

- Declaring a variable in the global scope with multiple template parameter variables would cause those template parameters to overwrite each other in memory. This fixes https://github.com/WerWolv/ImHex/issues/956.
- Type declarations would successfully parse and evaluate without enough template parameters for their struct, causing a null deref.
- The source location displayed when a variable already existed was always line 1, column 1.
- Any template parameter with the same name as a variable in the scope of the type declaration passing the parameter would cause an error, for example: `struct TestA<auto a> {}; struct TestB { u32 a = 0; TestA<a> test; }; TestB @ 0;`


lib/include/pl/core/ast/ast_node_type_decl.hpp
```diff
@@ -76,15 +76,46 @@ namespace pl::core::ast {
         [[nodiscard]] std::vector<std::shared_ptr<ptrn::Pattern>> createPatterns(Evaluator *evaluator) const override {
             evaluator->updateRuntime(this);
 
-            evaluator->pushTemplateParameters();
-            ON_SCOPE_EXIT { evaluator->popTemplateParameters(); };
+            std::vector<std::unique_ptr<ASTNodeLiteral>> templateParamLiterals(this->m_templateParameters.size());
 
-            for (const auto &templateParameter : this->m_templateParameters) {
+            // Get template parameter values before pushing a new template parameter scope so that variables from the parent scope are used before being overwritten.
+            for (size_t i = 0; i < this->m_templateParameters.size(); i++) {
```
WerWolv, line 82: This pattern here can generally be replaced with a for-each loop. `for (const auto &templateParameter : this->m_templateParameters)`

lib/include/pl/core/ast/ast_node_type_decl.hpp
```diff
@@ -76,15 +76,46 @@ namespace pl::core::ast {
...
+                evaluator->popTemplateParameters();
+            };
+
+            {
+                evaluator->pushSectionId(ptrn::Pattern::HeapSectionId);
+                ON_SCOPE_EXIT {
+                    evaluator->popSectionId();
+                };
+
+                for (size_t i = 0; i < this->m_templateParameters.size(); i++) {
```
WerWolv, line 106: Same here

lib/include/pl/core/ast/ast_node_type_decl.hpp
```diff
@@ -76,15 +76,46 @@ namespace pl::core::ast {
...
-            ON_SCOPE_EXIT { evaluator->popTemplateParameters(); };
+            std::vector<std::unique_ptr<ASTNodeLiteral>> templateParamLiterals(this->m_templateParameters.size());
 
-            for (const auto &templateParameter : this->m_templateParameters) {
+            // Get template parameter values before pushing a new template parameter scope so that variables from the parent scope are used before being overwritten.
+            for (size_t i = 0; i < this->m_templateParameters.size(); i++) {
+                auto &templateParameter = this->m_templateParameters[i];
                 if (auto lvalue = dynamic_cast<ASTNodeLValueAssignment *>(templateParameter.get())) {
+                    if (!lvalue->getRValue())
+                        err::E0003.throwError(fmt::format("There is no value for template parameter variable {}. This is a bug.", lvalue->getLValueName()), {}, this);
```
WerWolv, line 86: I'd rename this error to something like `No value set for non-type template parameter {}. This is a bug.`

WerWolv (changes requested):
Looks really clean all in all :) Thank you!

lib/include/pl/core/ast/ast_node_type_decl.hpp
Zaggy1024: This uses an index to look up the values in `m_templateParameters` and `templateParameterLiterals` at the same time, should I change this to use a for-each and an iterator declared outside the loop?

lib/include/pl/core/ast/ast_node_type_decl.hpp
Zaggy1024: This uses an index so that it can assign to the vector at the correct parameter index, leaving null pointers in all the other slots. I could change it to use for-each and an index outside the scope of the loop, though, up to you.

WerWolv (approved):

WerWolv:
I love this PR. This fixes basically all issues that existed with templates that I couldn't figure out myself. Thanks a lot!

PR #33, opened 2023-03-22
Create non-type template parameter variables with the correct type

Zaggy1024:
This fixes an issue where passing a pattern as a template parameter would result in the error message `Cannot cast from type 'Type' to type '???'.` due to not creating patterns for the specific custom type being assigned to the parameter.


WerWolv:
Thanks!

PR #34, opened 2023-03-22
Bitfields: Enable nesting, arrays, function calls, match, templates

Zaggy1024:
This also fixes the reading of little-endian bitfields, which should have been byte swapped. See https://github.com/WerWolv/ImHex/issues/981


lib/source/pl/lib/std/mem.cpp
```diff
@@ -128,6 +128,21 @@ namespace pl::lib::libstd::mem {
             });
 
 
+            /* current_bit_offset() */
+            runtime.addFunction(nsStdMem, "current_bit_offset", FunctionParameterCount::none(), [](Evaluator *ctx, auto params) -> std::optional<Token::Literal> {
+                wolv::util::unused(params);
+                return u128(ctx->getBitfieldBitOffset());
+            });
+
+            /* read_bits() */
```
WerWolv, line 137: Add function parameter names to this comment too

tests/include/test_patterns/test_pattern_bitfields.hpp
```diff
@@ -33,19 +40,27 @@ namespace pl::test {
...
+                bitfield NestedBitfield {
+                    nestedA : 4;
+                    nestedB : 4;
+                };
+
                 bitfield TestBitfield {
                     a : 4;
                     b : 4;
-                    c : 4;
+                    NestedBitfield c;
```
WerWolv, line 51: This test might benefit from an example for a bitfield array as well

lib/include/pl/core/ast/ast_node_bitfield_array_variable_decl.hpp
```diff
@@ -0,0 +1,206 @@
...
+                    err::E0001.throwError(fmt::format("Unexpected type of bitfield array size node."), {}, this);
+                }
+
+                auto limit = evaluator->getArrayLimit();
+                auto checkLimit = [&](auto count) {
+                    if (count > limit)
+                        err::E0007.throwError(fmt::format("Bitfield array grew past set limit of {}", limit), "If this is intended, try increasing the limit using '#pragma array_limit <new_limit>'.", this);
+                };
+
+                if (std::holds_alternative<u128>(boundsCondition))
```
WerWolv, line 135: You probably want to use `boundsCondition.toUnsigned()` here instead because it's valid to use signed numbers, chars or bools (or anything that can convert to a number really)

WerWolv (changes requested):

lib/include/pl/core/ast/ast_node_bitfield_array_variable_decl.hpp
Zaggy1024: The conversion happens above here at lines 118-122, I believe it should accept any of those types as that `std::visit()` call accepts the same types as `toUnsigned()`.

Zaggy1024:
Added bitfield arrays to the unit test, this should be good to go now!

WerWolv (commented):
Small style adjustments

WerWolv (approved):

PR #35, opened 2023-03-24
Fix refcounting of pattern-local variables

Zaggy1024:
This fixes accessing local variables from `ref auto` parameters, like in this example:

```cpp
#include <std/core.pat>

bitfield ZeroBit {
    zeroBit : 1;
    if (zeroBit) {
        break;
    }
} [[inline]];

bitfield ExpGolomb {
    ZeroBit leadingZeros[while (true)] [[sealed]];
    u32 value = (1 < 32) - 1;
    if ((std::core::member_count(leadingZeros) - 1) < 32) {
        valueBits : (std::core::member_count(leadingZeros) - 1) [[name("value")]];
        value = valueBits + (1 << (std::core::member_count(leadingZeros) - 1)) - 1;
    }
} [[format("get_exp_golomb_value"),transform("get_exp_golomb_value")]];
fn get_exp_golomb_value(ref auto exp_golomb) {
    return exp_golomb.value;
};

ExpGolomb e @ 0;
```

Previously, the ref count was starting at 0, decrementing due to an unused pattern in `ASTNodeVariableDecl::createPatterns` and wrapping to the maximum value, then incrementing on entering the formatter function, causing the evaluator to see a 0 ref count and delete the variable's memory.

### DIFF
--- a/lib/include/pl/core/ast/ast_node_variable_decl.hpp
+++ b/lib/include/pl/core/ast/ast_node_variable_decl.hpp
@@ -75,31 +75,31 @@ namespace pl::core::ast {
                     err::E0005.throwError(fmt::format("Cannot place variable '{}' at out of bounds address 0x{:08X}", this->m_name, evaluator->dataOffset()), { }, this);
             }
 
-            auto patterns = this->m_type->createPatterns(evaluator);
-            if (patterns.empty())
-                err::E0005.throwError("'auto' can only be used with parameters.", { }, this);
-
-            auto &pattern = patterns.front();
-            if (this->m_placementOffset != nullptr && dynamic_cast<ptrn::PatternString*>(pattern.get()) != nullptr)
-                err::E0005.throwError(fmt::format("Variables of type 'str' cannot be placed in memory.", this->m_name), { }, this);
-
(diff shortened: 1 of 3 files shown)


Zaggy1024:
@WerWolv I'm not sure if this is the right solution. It seems like it would be better if patterns would actually require a section ID as a constructor parameter and not have a setter for the section ID, so that the ref counting can always be reliable. However, I don't know if anything currently depends on setting the section ID dynamically, so I'll have to look into that later if you think it's worth doing.

Leaving this as a draft until then.

Zaggy1024:
Updated this with hopefully a much more robust solution. I figured it would be a huge amount of work to make sections and offsets be set on pattern creation, so instead I made `Pattern::setSection()` and `Pattern::setOffset()` call `Evaluator::destroyPattern()` and `Evaluator::createPattern()` respectively before and after setting the values. This allows the refcounts to remain accurate, at least in my preliminary testing.

I also made it an error to refcount a deleted local, and made `Evaluator` clean up its patterns before clearing the local storage so that all locals are deleted properly first.

WerWolv:
Thank you!

PR #36, opened 2023-03-25
Allow match statements' cases to specify extra conditions

Zaggy1024:
This adds parsing of extra conditional statements on match cases similar to how Rust allows:
```rust
u32 conditionallyRelevant;

struct MatchConditional {
    u8 value;

    match (value) {
        (0): u32 item;
        (1) if (conditionallyRelevant == 0): u64 item;
        (_): u16 item;
    }
}
```

This could also be achieved by the following:
```rust
u32 conditionallyRelevant;

struct MatchConditional {
    u8 value;

    match (value, conditionallyRelevant) {
        (0, _): u32 item;
        (1, 0): u64 item;
        (_, _): u16 item;
    }
}
```
but I think it helps with readability in otherwise long match statements that have a few cases where some other variable is relevant.

This PR also includes some small improvements to the match parsing code.

### DIFF
--- a/lib/include/pl/core/parser.hpp
+++ b/lib/include/pl/core/parser.hpp
@@ -136,9 +136,7 @@ namespace pl::core {
         std::unique_ptr<ast::ASTNode> parseFunctionVariableAssignment(const std::string &lvalue);
         std::unique_ptr<ast::ASTNode> parseFunctionVariableCompoundAssignment(const std::string &lvalue);
         std::unique_ptr<ast::ASTNode> parseFunctionControlFlowStatement();
-        std::vector<std::unique_ptr<ast::ASTNode>> parseStatementBody();
-        std::unique_ptr<ast::ASTNode> parseFunctionConditional();
-        std::unique_ptr<ast::ASTNode> parseFunctionMatch();
+        std::vector<std::unique_ptr<ast::ASTNode>> parseStatementBody(const std::function<std::unique_ptr<ast::ASTNode>()> &memberParser);
         std::unique_ptr<ast::ASTNode> parseFunctionWhileLoop();
         std::unique_ptr<ast::ASTNode> parseFunctionForLoop();
 
--- a/lib/source/pl/core/parser.cpp
+++ b/lib/source/pl/core/parser.cpp
@@ -549,10 +549,10 @@ namespace pl::core {
         else if (MATCHES(oneOf(tkn::Keyword::Return, tkn::Keyword::Break, tkn::Keyword::Continue)))
             statement = parseFunctionControlFlowStatement();
         else if (MATCHES(sequence(tkn::Keyword::If))) {
-            statement      = parseFunctionConditional();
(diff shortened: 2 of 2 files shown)


WerWolv:
Thank you!
I'm not sure I'm 100% comfortable with this syntax though. Both examples here look almost the same but have a vastly different outcome:
```rust
u32 conditionallyRelevant;

struct MatchConditional {
    u8 value;

    match (value) {
        (0): u32 item;
        (1) if (conditionallyRelevant == 0): u64 item;
        (_): u16 item;
    }
}
```

```rust
u32 conditionallyRelevant;

struct MatchConditional {
    u8 value;

    match (value) {
        (0): u32 item;
        (1): if (conditionallyRelevant == 0) u64 item;
        (_): u16 item;
    }
}
```

Maybe moving the if in front of the case would improve this? I'm not sure but I'm open for discussion

Zaggy1024:
I think when reading the code it makes more sense to me to have the extra condition after the case, since it keeps things aligned, and it also matches the way Rust does it:

```rust
let num = Some(4);

match num {
    Some(x) if x % 2 == 0 => println!("The number {} is even", x),
    Some(x) => println!("The number {} is odd", x),
    None => (),
}
```

The difference I see is that the `:` separator is a lot less visible than `=>`, which made it a lot more obvious in Rust that it is part of the match statement and not a separate condition. However, I think I would prefer either this syntax or just leaving people to use the multi-match syntax, which is also probably fine.

KillyMXI:
Rust example above shows the main point for extra conditions - extra condition applied to a variable within the matched type.
Is the same applicable for pattern language?
Otherwise, if only external variables can be checked, then `match (value, conditionallyRelevant)` makes more sense to me.

Zaggy1024:
Good point, although I think that in certain cases it could get messy including multiple variables in match cases. Say you have one variable that is always relevant, then several others that are only relevant to specific values of the first?

As it is, this would require something like the following:

```rust
u32 a;
u32 b;
u32 c;

struct MatchConditional {
    u8 value;

    match (value, a, b, c) {
        (0, _, _, _): ...
        (1, _, 0, _): ...
        (2, _, _, _): ...
        (3, _, _, _): ...
        (4, 0, _, _): ...
        (5, _, _, 0): ...
        (_, _, _, _): ...
    }
}
```

Perhaps there's an argument to be made that this should be done with `if`/`else if` statements instead, but I think match statements are a lot cleaner when you need to base this type of thing off one variable's value.

KillyMXI:
I actually kinda like the table look, even if this sparse - I can argue it is more readable compared to extra conditions.

Zaggy1024:
Perhaps that was a bad example too though, I'm thinking of when either the values' widths don't line up (say with enums where names can be quite long), or if the match has a lot of cases. In those cases, having to refer back to the match's parameters feels like it would be a bit of a pain.

To be fair, though, this is a theoretical situation, as of yet I don't have such a match statement :^)

Zaggy1024:
Closing this for now to open a new PR with separate improvement.

PR #39, opened 2023-03-26
Refer to the correct column when a bitfield body doesn't start with '{'

Zaggy1024:
Previously, this would point to the bitfield name instead of the token that caused the error.


WerWolv:
Thank you!

PR #50, opened 2023-04-04
Don't set bitfields members' offsets that are in a different section

Zaggy1024:
I apparently forgot to put this into my previous PRs. :( This prevents the pattern-local vectors from growing out of control within bitfields.


WerWolv:
Thank you!

PR #53, opened 2023-05-18
Changes to fix issues editing floating point arrays placed on sections.

paxcut:
Issues 1009 WerWolv/ImHex/issues/1009  and 1038 WerWolv/ImHex/issues/1038 deal with problems setting values to  arrays of floating point values that are placed on sections.
The arrays were being resized to size one deleting all but the first and last values (issue 1009) and also resetting the section stack (issue 1038). This PR aims to fix both issues while leaving all other behavior intact.


WerWolv:
Thank you!

PR #59, opened 2023-06-16
Don't rename variables that are referenced by transform functions

Zaggy1024:
Function calls were previously changed to use a shared pointer to patterns that are referenced by parameters. This meant that those patterns would be renamed to the parameter's variable name, which needed to be undone by `ASTNodeFunctionCall`. However, this wasn't mirrored in the call to transform functions in `ASTNodeRValue`, meaning that those could cause variable names to be incorrect.


WerWolv:
Thanks!

PR #65, opened 2023-08-16
Fixed setDataSource seeking the wrong offset

LennardF1989:
On Unix (or at least Ubuntu), the original line would cause baseAddress to become something else than its intended value after leaving the executePattern method (eg. when the formatter uses the same runtime). I suspected a scope issue, so I changed the baseAddress to be a copy, rather than a reference and this fixed the issue.

I didn't want to change too much, but an alternative method that could also work is adding a getDataBaseAddress to the PatternLanguage-class and use that in the lambda instead.

### DIFF
--- a/cli/source/helpers/utils.cpp
+++ b/cli/source/helpers/utils.cpp
@@ -24,7 +24,8 @@ namespace pl::cli {
         for (const auto &define : defines)
             runtime.addDefine(define);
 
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

rvanlaar:
Great work. Thanks

PR #72, opened 2023-11-16
feat: make --type optional + minor improvements

iTrooz:
Commit descriptions should be self-explanatory

### DIFF
--- a/cli/source/subcommands/format.cpp
+++ b/cli/source/subcommands/format.cpp
@@ -120,6 +120,7 @@ namespace pl::cli::sub {
             }
 
             outputFile.writeVector(result);
+            ::fmt::print("Wrote pattern data to {}\n", outputFilePath.string());
         });
     }
 
--- a/cli/source/subcommands/format.cpp
+++ b/cli/source/subcommands/format.cpp
@@ -23,19 +23,19 @@ namespace pl::cli::sub {
         static bool metaInformation = false;
         static u64 baseAddress = 0x00;
 
-        auto subcommand = app->add_subcommand("format");
+        auto subcommand = app->add_subcommand("format", "Executes the given pattern on the given file, and output the pattern data");
 
         // Add command line arguments
-        subcommand->add_option("-i,--input,INPUT_FILE", inputFilePath, "Input file")->required()->check(CLI::ExistingFile);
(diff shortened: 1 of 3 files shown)


cli/source/subcommands/format.cpp
```diff
@@ -23,19 +23,19 @@ namespace pl::cli::sub {
         static bool metaInformation = false;
         static u64 baseAddress = 0x00;
 
-        auto subcommand = app->add_subcommand("format");
+        auto subcommand = app->add_subcommand("format", "Executes the given pattern on the given file, and output the pattern data");
```
WerWolv, line 26: I'd mention here that it outputs the data as some other format

cli/source/subcommands/format.cpp
```diff
@@ -23,19 +23,19 @@ namespace pl::cli::sub {
...
 
         // Add command line arguments
-        subcommand->add_option("-i,--input,INPUT_FILE", inputFilePath, "Input file")->required()->check(CLI::ExistingFile);
+        subcommand->add_option("-i,--input,INPUT_FILE", inputFilePath, "Input file to parse")->required()->check(CLI::ExistingFile);
```
WerWolv, line 29: Does it really parse the binary file? Isn't there a better word for this?

cli/source/subcommands/run.cpp
```diff
@@ -22,7 +22,7 @@ namespace pl::cli::sub {
         auto subcommand = app->add_subcommand("run");
 
         // Add command line arguments
-        subcommand->add_option("-i,--input,INPUT_FILE", inputFilePath, "Input file")->check(CLI::ExistingFile);
+        subcommand->add_option("-i,--input,INPUT_FILE", inputFilePath, "Input file to parse")->check(CLI::ExistingFile);
```
WerWolv, line 25: Same as above

cli/source/subcommands/format.cpp
iTrooz: I think "parse" is fine, but we could also say "input file to extract data from"

WerWolv: I like that more tbh

iTrooz: Okay, i'll change it !

PR #75, opened 2024-01-16
fix: Some errors related to preprocessor and lexer

lenrik1589:
Doc comments (both single line and multiline) now act as comments during preprocessing stage, but are kept for later use by lexer, which now supports single line doc comments now too.
Improved support for older macs! /hj (files with `\r` as line ending, are now handled by preprocessor by internally converting `\r` and `\r\n` line endings into `\n`).
Using multiline comments no longer mis-aligns execution errors.
Preprocessor directives are now listed in full instead of only "include" "define" and "pragma".
Not having an argument after directives with explicit arguments no longer throws cryptic error if there is no space, and throws an error if there is only a space;

### DIFF
--- a/lib/include/pl/pattern_language.hpp
+++ b/lib/include/pl/pattern_language.hpp
@@ -81,7 +81,7 @@ namespace pl {
          * @param checkResult Whether to check the result of the execution
          * @return True if the execution was successful, false otherwise
          */
-        [[nodiscard]] bool executeString(std::string code, const std::map<std::string, core::Token::Literal> &envVars = {}, const std::map<std::string, core::Token::Literal> &inVariables = {}, bool checkResult = true);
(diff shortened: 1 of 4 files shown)


WerWolv:
Thank you! Sorry, I didn't see your PR until now.
I think this has been mostly superseded by #77 now. What do you say @paxcut?

paxcut:
most definitely. The code is still using pre-refactored code so it can't be used as is, but I made sure (or at least tried ) to duplicate some of the fixes made here like handling /**/ and #define (nothing after last 'e')

lenrik1589:
> Thank you! Sorry, I didn't see your PR until now.

Sorry for not responding, had life™, there was a discussion in the discord @ the time i posted the PR, you were in it as well, but it was in fact intended as a temporary fix for the old system, if the new system integrates these changes it's fine by me

paxcut:
I tried to implement all the changes you made and I did some testing for them. If you have a set of tests to run on the latest nightly it would be great if you could run them real quick in case something was missing.I am now finishing last details on a full syntax highlighter for the pattern editor which has to do with some of the issues addressed in this pr like comment delimiters under some circumstances so I could add more changes if needed.

PR #77, opened 2024-02-10
Added #undef and some fixes to preprocessor.

paxcut:
feat: added #undef preprocessor directive.  
fix: missing char in message of lexer.
fix: preprocessor no longer does replacements inside strings or doc comments. Also defines are only applied after their position in the code instead of globally. Redefinition of macros is allowed but a warning message is issued. Defines are only allowed to use non-empty and valid identifiers as macro names. 
fix: problem that directives without complete syntax did not generate error messages. (as in #define with nothing after last e)
fix: some combinations  of comments were not handled correctly for example /**/ was parsed as a beginning doc comment.
fix: nested #ifdef directives were not parsed correctly producing errors elsewhere.
fix: The deletion of comments by preprocessor resulted in wrong column numbers being displayed in error messages.
fix: two consecutive directives with no space between them caused the failure to parse the second directive.

Needs work: if an error occurs inside an included file, its location is reported as it is in the preprocessed code. Since the entire file is stored in a single line to preserve line numbers in including pattern this results in making it hard to track the error. The reported column number needs to be converted to the correct position of the error in the file. Better yet would be to add references to the file being included so they can be parsed in their original form and generate better errors.

The code modifications try to adhere to the spirit of the recent code refactoring.  Adding new code is chosen over modifying existing code whenever possible within reasonable limits. Please review at your earliest convenience.

### DIFF
--- a/lib/include/pl/api.hpp
+++ b/lib/include/pl/api.hpp
@@ -28,7 +28,7 @@ namespace pl::api {
      */
     using PragmaHandler = std::function<bool(PatternLanguage&, const std::string &)>;
 
-    using DirectiveHandler = std::function<void(core::Preprocessor*, unsigned)>;
+    using DirectiveHandler = std::function<void(core::Preprocessor*, u32)>;
 
     using Resolver = std::function<hlp::Result<Source*, std::string>(const std::string&)>;
 
--- a/lib/include/pl/api.hpp
+++ b/lib/include/pl/api.hpp
@@ -28,7 +28,7 @@ namespace pl::api {
      */
     using PragmaHandler = std::function<bool(PatternLanguage&, const std::string &)>;
 
-    using DirectiveHandler = std::function<void(core::Preprocessor*)>;
+    using DirectiveHandler = std::function<void(core::Preprocessor*, unsigned)>;
 
     using Resolver = std::function<hlp::Result<Source*, std::string>(const std::string&)>;
 
--- a/lib/include/pl/core/preprocessor.hpp
+++ b/lib/include/pl/core/preprocessor.hpp
@@ -5,6 +5,7 @@
 #include <set>
 #include <string>
 #include <unordered_map>
+#include <atomic>
 
 #include <pl/api.hpp>
 #include <pl/helpers/types.hpp>
@@ -23,7 +24,7 @@ namespace pl::core {
 
         ~Preprocessor() override = default;
 
-        hlp::CompileResult<std::string> preprocess(PatternLanguage *runtime, api::Source* source, bool initialRun = true);
+        hlp::CompileResult<std::vector<pl::core::Token>> preprocess(PatternLanguage *runtime, api::Source* source, bool initialRun = true);
 
         void addDefine(const std::string &name, const std::string &value = "");
         void addPragmaHandler(const std::string &pragmaType, const api::PragmaHandler &handler);
@@ -31,12 +32,14 @@ namespace pl::core {
         void removePragmaHandler(const std::string &pragmaType);
         void removeDirectiveHandler(const std::string &directiveType);
 
-        std::optional<std::string> getDirectiveValue(bool allowWhitespace = false);
-        void replaceSkippingStrings(std::string &input, const std::string &find, const std::string &replace);
-        void saveDocComment();
+        std::optional<pl::core::Token> getDirectiveValue(unsigned line);
 
         [[nodiscard]] bool shouldOnlyIncludeOnce() const {
             return this->m_onlyIncludeOnce;
+        }
+
+        bool isInitialized() const {
+            return m_initialized;
         }
 
         void setResolver(const api::Resolver& resolvers) {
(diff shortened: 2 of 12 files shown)


lib/include/pl/core/preprocessor.hpp
```diff
@@ -23,70 +24,71 @@ namespace pl::core {
...
+        hlp::CompileResult<std::vector<pl::core::Token>> preprocess(PatternLanguage *runtime, api::Source* source, bool initialRun = true);
 
         void addDefine(const std::string &name, const std::string &value = "");
         void addPragmaHandler(const std::string &pragmaType, const api::PragmaHandler &handler);
         void addDirectiveHandler(const std::string &directiveType, const api::DirectiveHandler &handler);
         void removePragmaHandler(const std::string &pragmaType);
         void removeDirectiveHandler(const std::string &directiveType);
 
-        std::optional<std::string> getDirectiveValue(bool allowWhitespace = false);
+        std::optional<pl::core::Token> getDirectiveValue(unsigned line);
```
WerWolv, line 35: Generally use sized types like `u32` instead of `unsigned` or `int` and co.

lib/source/pl/core/lexer.cpp
```diff
@@ -105,6 +105,67 @@ namespace pl::core {
         return c;
     }
 
+    std::optional<Token> Lexer::parseDirectiveName(const std::string_view &identifier) {
+        auto directives = Token::Directives();
```
WerWolv, line 109: This can probably be a const reference to avoid copying the entire data

lib/source/pl/core/preprocessor.cpp
```diff
@@ -38,90 +48,83 @@ namespace pl::core {
...
 
     void Preprocessor::processIfDef(const bool add) {
         // find the next #endif
         const Location start = location();
         u32 depth = 1;
-        while(peek() != 0 && depth > 0) {
-            if(m_code.substr(m_offset, 6) == "#endif" && !m_inString && m_startOfLine) {
-                m_offset += 6;
+        while(peek() != "EOF" && depth > 0) {
+            if(m_token->getFormattedValue() == "#endif") {
```
WerWolv, line 113: Instead of comparing the formatted value here, would it be possible to compare the token itself like it's done in other functions?

lib/source/pl/core/preprocessor.cpp
```diff
@@ -131,77 +134,109 @@ namespace pl::core {
...
             errorDesc("No file to include given in #include directive.", "A #include directive expects a path to a file: #include \"path/to/file\" or #include <path/to/file>.");
             return;
         }
 
-        if (includeFile->starts_with('"') && includeFile->ends_with('"'))
-            ; // Parsed path wrapped in ""
-        else if (includeFile->starts_with('<') && includeFile->ends_with('>'))
-            ; // Parsed path wrapped in <>
-        else {
+        if ((!includeFile.starts_with('"') || !includeFile.ends_with('"')) &&
```
WerWolv, line 233: I feel like this could be simplified using de-morgan

WerWolv (changes requested):

lib/include/pl/core/preprocessor.hpp
```diff
@@ -23,70 +24,71 @@ namespace pl::core {
...
 
-        size_t m_offset = 0;
-        u32 m_lineNumber = 1;
-        u32 m_lineBeginOffset = 0;
-        bool m_inString = false;
-        bool m_startOfLine = true;
-        std::string m_output;
+        std::vector<pl::core::Token> m_keys;
+        std::atomic<bool> m_initialized = false;
+        __gnu_cxx::__normal_iterator<pl::core::Token *, std::vector<pl::core::Token>> m_token;
```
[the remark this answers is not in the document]
paxcut: I just want an iterator that I can increment, decrement and de-reference that lets me access the vector of tokens. I tried using the vector iterator but there were problems with Token constructor. I tried using an std::iterator but gcc only implements it that way.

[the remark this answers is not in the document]
jumanji144: Well we can't use `__gnu_cxx::__normal_iterator` as that is a internal implementation type, using this is considered undefined behaviour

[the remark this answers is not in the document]
WerWolv: You should be able to do `std::vector<pl::core::Token>::iterator` or `decltype(m_keys)::iterator`.
I'm doing something similar in the parser already, you can check there

lib/source/pl/core/preprocessor.cpp
paxcut: yes, it is more efficient that way. In fact strings shouldn't even be used at this point since the token enums can do a much better job. For example the directive handler map can be from Token::Directive to function and the same for the rest.

lib/include/pl/core/preprocessor.hpp
paxcut: I changed all the unsigned types to u32. Let me know if I missed any.

lib/include/pl/core/preprocessor.hpp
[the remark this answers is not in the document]
paxcut: Ok, vector iterator seems to be working now so I switched to it.

lib/include/pl/api.hpp
```diff
@@ -28,7 +28,7 @@ namespace pl::api {
      */
     using PragmaHandler = std::function<bool(PatternLanguage&, const std::string &)>;
 
-    using DirectiveHandler = std::function<void(core::Preprocessor*)>;
+    using DirectiveHandler = std::function<void(core::Preprocessor*, unsigned)>;
```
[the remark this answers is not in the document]
paxcut: that was changed as well.

lib/source/pl/core/lexer.cpp
paxcut: I added a bunch of const refs but that one escaped. Not anymore.

lib/source/pl/core/preprocessor.cpp
paxcut: I removed all the string comparisons and switched to tokens instead even for the directive handler map.

lib/source/pl/core/preprocessor.cpp
paxcut: I made this statement the simplest I could.

PR #92, opened 2024-04-30
evaluator: Fixed section pointers in ifs and try/catch

zziger:
Same as #90 but for pointers located in conditional or try/catch statements


WerWolv:
Thank you!

PR #93, opened 2024-04-30
preprocessor/test: Add test for pragmas + make executeString() fail on parsing error + fix including an already-included file

iTrooz:


iTrooz:
Some context is somewhat here: https://discord.com/channels/789833418631675954/789840633414025246/1234255400333742181

PR #115, opened 2024-08-02
It is finally time to commit the entire syntax highlighting code and first is always the pattern language

paxcut:
There are also a number of fixes and improvements as listed next.

fix: problem with having no spaces between `ifdef` and `endif` directives. renamed `patternLocalvariable` to `localVariable` and `patternPlacedVariable` to `calculatedPointer` Also introduced a view type so now we have 4 struct types, 3 function types and 2 global types.
fix: global arrays with new initialization syntax showed as unknown types.
Syntax highlighting doesn't use the `ast` anymore.
fix: preprocessor directives cannot have code on the same line they are except for comments. Any extra tokens located after a fully parsed  directive are simply ignored. Earlier versions of the highlighting code used two sets of tokens. It needed to have access to the token sequence as it is before preprocessing so that it contains all comments and directives. It also needed access to the sequence that was created by the parser because it contains all the identifier definitions types. Now the two are merged together in the preprocessor making highlighting simpler


paxcut:
most of the fixes that were included here have already been merged as separate prs already. there is some code that cannot be broken apart that will be submitted soon that deals with the syntax highlights only. it also improves the code readability and makes bugs easier to track.  This has been left as a draft in case i missed something or cannot get something working that worked here but it is so old the it isnt worth much anymore.

PR #120, opened 2024-08-28
fix: Double template treated as right shift

paxcut:
A template of the form A<B<count>> a@0; is being parsed as count >> a.

 The fix consists of the same method used to disambiguate the closing template operator from the greater than comparison operator.


paxcut:
Fixed already in its own commit so closing this

PR #122, opened 2024-08-28
fix: Using the windows path delimiter \ results in error.

paxcut:
This can be an issue if you are copying the path directly from explorer. Currently, the old highlighter code changes the color of the char but that will change with the new code. The fix was to add a parameter to the function that processes the directives in the lexer to allow it to skip the escape check.


WerWolv:
Personally, I hate having windows paths in source files because it basically implies that the file will only ever work on Windows. Using the standard unix path separators work everywhere so I don't think this is a good idea to add. Besides, if we support \, we'll also have to support things like Japanese paths which uses ¥ as the path separator.

PR #128, opened 2024-09-11
fix: Broken #pragma once on import

zziger:
## Issue description

`import`ing a `#pragma once` marked file multiple times results in multiple evaluation. (E.g. having a variable will throw an error for multiple definitions)

## Issue cause
Already included files marked with #pragma once are stored in a std::set with key of custom type OnceIncludePair. The type overrides spaceship operator for map/set comparison, but uses data from api::Source pointer.  The api::Source pointer is a pointer to a map element, which is stored to keep one instance for all loaded files from the same path.
The issue hides in the fact that default resolver [stores the instance through insert_or_assign](https://github.com/WerWolv/PatternLanguage/blob/188a8089a8f23d0cdacc0f6c38384e4f8fe93ef1/lib/source/pl/core/resolver.cpp#L25) which replaces the old data at this pointer with the new one, with a different ID. Since old comparison used IDs when added to a onceIncluded set, and the ID is now different, set order is not valid anymore, breaking set::contains and potentially other set operations.

## Proposed solution
This PR compares api::Source pointers directly, instead of IDs inside of them. This ensures that no data inside will change (and therefore won't break set/map structure), while still keeping the functionality, since the pointer doesn't change.

### DIFF
--- a/lib/include/pl/core/parser_manager.hpp
+++ b/lib/include/pl/core/parser_manager.hpp
@@ -45,7 +45,9 @@ namespace pl::core {
             std::string alias;
 
             auto operator<=>(const OnceIncludePair& other) const {
-                return std::tie(*this->source, this->alias) <=> std::tie(*other.source, other.alias);
+                auto sourcePtr = reinterpret_cast<uintptr_t>(source);
+                auto otherSourcePtr = reinterpret_cast<uintptr_t>(other.source);
+                return std::tie(sourcePtr, this->alias) <=> std::tie(otherSourcePtr, other.alias);
             }
         };
 


jumanji144:
Hello, thanks for the fix. I believe the reason we added the ID was because we deemed pointer comparison unsafe and IDs to be more unique, and therefore equivalent.

But if this proves to be equivalent and fix the proposed bug, then it should be good

paxcut:
Although the error is being described in detail, there is no example that exposes the error that can be tested for reproduction. Files imported that have pragma are re-imported continuously as dependencies run several levels deep. One obvious test is to import the same file twice and that didn't show any problems either. I'm not saying there is no bug,  just that I can't one and nothing indicating how to trigger it is being described.

jumanji144:
What about the failing unit tests?

paxcut:
failing unit test has been happening for a long time now.

paxcut:
for example check [here](https://discord.com/channels/789833418631675954/789840633414025246/1278357696529174528)

zziger:
@paxcut simple repro setup:

file1.pat
```rs
#pragma once
namespace something {
    auto SOME_CONST = 15;
}
```

Pattern editor window
```rs
import file1;
import file1;
```

Result
```
E: [ Stack Trace ]
E: /Users/zziger/Source/formats-imhex/patterns/file1.pat:4:21
E: 4 |     auto SOME_CONST = 15;
E:                         ^


E: runtime error: Variable with name 'SOME_CONST' already exists in this scope.
E:   -->   in /Users/zziger/Source/formats-imhex/patterns/file1.pat:4:5
E: 4 |     auto SOME_CONST = 15;
E:         ^^^^
```

paxcut:
Yes, that's a situation one rarely encounters which would explain why it wasn't detected earlier. To be sure, the brief description given on the original post is not enough to describe the error. I seems it requires that the imported file is not part of the include library that is provided. For some reason those headers don't have problems with `#pragma once`

paxcut:
the reason the error occurs is because when it uses this code to check if header was already included, it still adds the AST nodes when there are no types defined in the children (imports of imports) of the imported file.
```cpp
    if (m_onceIncluded.contains( key )) {
        const auto& types = m_parsedTypes[key];
        if (!types.empty())
            return Result::good({ {}, types });
    }
```
The fix (and it works when tested) is to return a result even if types is empty to avoid returning ast nodes further down.
```cpp
    if (m_onceIncluded.contains( key )) {
        const auto& types = m_parsedTypes[key];
        return Result::good({ {}, types });
    }
```

WerWolv:
This has been superseded by #133, correct? @paxcut

paxcut:
yes, it is explained there in detail

PR #129, opened 2024-09-16
Fix native build support for macos

aedrax:
# Summary
This pull request addresses build issues related to deprecated functions and improves compatibility with different C++ compilers. The changes primarily include modifying the CMakeLists.txt to suppress deprecated warnings for both Clang and AppleClang compilers and updating the project source code to resolve errors caused by constructor mismatches.

I'm on an M1 MacBook Pro
version 14.6.1 (23G93)

and I'm building this way:
```
cmake -B build
cmake --build build -- -j$(sysctl -n hw.ncpu)
```

### 1. Suppress Deprecated Warnings in CMake:
- Enhanced `CMakeLists.txt` to handle both Clang and AppleClang compilers by adding a condition to suppress deprecated warnings.

### 2. Fixed Constructor Mismatch in `Scope` Struct:
- Updated the `Scope` struct constructor to match the parameters provided during its instantiation.
- Removed inappropriate usage of `std::optional<std::nullopt_t>`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-stringop-overflow -Wno-dangling-reference -fexceptions -frtti")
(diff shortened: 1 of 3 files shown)


joshleaves:
Thanks for this, works on my machine too. Hope it gets merged to master soon.

aedrax:
> Not sure about second part of the commit, maybe it should go into its own pull request to separate both concerns?

Second part was required for it to build on Macos, otherwise I'd agree that it should be a separate Pull Request

PR #131, opened 2024-09-30
fix: breakpoints become disabled after first hit.

paxcut:
If you set a breakpoint the first time you run the pattern it gets hit. After that subsequent runs will fail to hit the breakpoint.

This bug was caused by two factors.
- When breakpoint hits evaluation stops before the line with the breakpoint is evaluated. In order to avoid hitting the breakpoint again when evaluation is continues the breakpoint needs to be disabled. Once the line is evaluated the breakpoint needs to be enabled again, but the previous code didn't do that.
- Detecting if the line was being evaluated used insufficient information to perform said detection. Because of the way AST nodes are assigned coordinates, Nodes that lie in other lines iof source code may and up sharing the same line number as the Nodes in the breakpoint. To tell them apart one must also use columns

Both problems were resolved. When evaluation of another line occurs the breakpoint is re-enabled. Note that at this point the full evaluation of the line may not have finish yet. This is of no concern as the second problem prevents triggering the breakpoint erroneously. By using the column of the position of the node we ensure that only valid evaluations of the node trigger the breakpoints.


lib/source/pl/core/evaluator.cpp
```diff
@@ -1006,6 +1007,18 @@ namespace pl::core {
         if (this->isDebugModeEnabled())
             this->m_console.log(LogConsole::Level::Debug, fmt::format("Base Pattern size: 0x{:02X} bytes", sizeof(ptrn::Pattern)));
 
+        this->m_sourceLineLength.clear();
+        for (auto &topLevelNode : ast) {
+            if (topLevelNode->getLocation().source->source == "<Source Code>") {
```
WerWolv, line 1012: Instead of checking for `<Source Code>` manually here (which does not always mean it's the top level file), instead use the `source->mainSource` member

WerWolv (changes requested):

lib/source/pl/core/evaluator.cpp
```diff
@@ -1006,6 +1007,18 @@ namespace pl::core {
         if (this->isDebugModeEnabled())
             this->m_console.log(LogConsole::Level::Debug, fmt::format("Base Pattern size: 0x{:02X} bytes", sizeof(ptrn::Pattern)));
 
+        this->m_sourceLineLength.clear();
+        for (auto &topLevelNode : ast) {
+            if (topLevelNode->getLocation().source->mainSource) {
```
paxcut, line 1012: I added the suggested change (clion kept telling me variable didn't exist)

lib/source/pl/core/evaluator.cpp
paxcut: OK, change was made as requested change. It seems like there may be edge cases when <Source Code> is not what is in the pattern editor because from what I can tell it should be but I take your word that mainSource will get it 100% of the time.

PR #142, opened 2024-12-07
feat: add `-P` flag to `info` subcommand to get metadata about multiple patterns at the same time + remove `--type` flag

iTrooz:


iTrooz:
Expected use case: use in https://github.com/werWolv/WerwolvAPI to parse patterns metadata quicker.
From a test I've done, this reduce the time to parse all patterns metadata from 5.9s to 2.8s

PR #145, opened 2024-12-28
parser: Add support for parsing RValue member/struct assignments in struct

### LINKED ISSUE #87: Struct member access doesn't work on the left-hand side of +=, nor the left-hand side within patterns
Using the version of Pattern Language in ImHex v1.33.2, if `abc` is a struct, then `abc.var` is a little picky on the left-hand side of assignments in the global scope, and very picky on the left-hand side of assignments within structs (regardless of whether `abc` is a global struct or a member of the struct currently being parsed).
(This might explain half of https://github.com/WerWolv/ImHex/issues/1599 ?)

\
Note the `//This does not work` comments in...
```rust
import std.io;

struct GlobalStruct {
    u8 var = 10;
};
GlobalStruct abc;

abc.var = 15; //This works
//abc.var += 5; //This does not work
abc.var = abc.var + 5; //This works
std::print("{}",abc.var+2); //This works
u8 global_var1 = abc.var + 5 [[export]]; //This works
u8 global_var2 = 0 [[comment(abc.var+3),export]]; //This works

struct MyStruct {
    //abc.var = 50; //This does not work
    //abc.var += 5; //This does not work
    //abc.var = abc.var + 5; //This does not work
    std::print("{}",abc.var+2); //This works
    u8 struct_var1 = abc.var + 5 [[export]]; //This works
    u8 struct_var2 [[comment(abc.var+3),export]]; //This works

    struct_var1 = 70; //This works
    //this.struct_var1 = 75; //This does not work
    u8 struct_var3 = this.struct_var1 + 4 [[export]]; //This works, meaning that "this.struct_var1" above isn't incorrect
};

MyStruct data @ $;
```

AkiSakurai:
Partially fix #87

### DIFF
--- a/lib/include/pl/core/parser.hpp
+++ b/lib/include/pl/core/parser.hpp
@@ -145,6 +145,7 @@ namespace pl::core {
         hlp::safe_unique_ptr<ast::ASTNode> parseScopeResolution();
         hlp::safe_unique_ptr<ast::ASTNode> parseRValue();
         hlp::safe_unique_ptr<ast::ASTNode> parseRValue(ast::ASTNodeRValue::Path &path);
+        hlp::safe_unique_ptr<ast::ASTNode> parseRValueAssignment();
         hlp::safe_unique_ptr<ast::ASTNode> parseUserDefinedLiteral(hlp::safe_unique_ptr<ast::ASTNode> &&literal);
         hlp::safe_unique_ptr<ast::ASTNode> parseFactor();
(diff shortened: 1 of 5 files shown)


WerWolv:
Thank you very much! Code looks good to me, I'll play around with it a bit soon

paxcut:
> Partially fix https://github.com/WerWolv/PatternLanguage/issues/87

I can't help but wonder why this was said to partially fix issue #87. What aspect of that issue is not covered?

PR #147, opened 2024-12-30
Prioritise local parameters over template parameters in AST node rvalue evaluation

### LINKED ISSUE #146: Local variable is shadowed by a template parameter of an ancestor
``` c++
import std.io;

struct inner 
{
  u32 size = 42;
  std::print("size:{}\nthis.size:{}", size, this.size); 
};

struct outer<auto size> 
{
    inner inner;
};

outer<24> outer;
// I: size:24
// I: this.size:42
```

AkiSakurai:
- fix #146

Unsure if it will break the existing pattern. Is there any case where the original behavior is preferred?

### DIFF
--- a/lib/source/pl/core/ast/ast_node_rvalue.cpp
+++ b/lib/source/pl/core/ast/ast_node_rvalue.cpp
@@ -163,13 +163,13 @@ namespace pl::core::ast {
         }
 
         {
-            const auto &currScope = evaluator->getScope(0);
-            std::copy(currScope.scope->begin(), currScope.scope->end(), std::back_inserter(searchScope));
+            const auto &templateParameters = evaluator->getTemplateParameters();
+            std::copy(templateParameters.begin(), templateParameters.end(), std::back_inserter(searchScope));
         }
 
         {
-            const auto &templateParameters = evaluator->getTemplateParameters();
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

I don't believe the original behaviour was correct. You can try running the tests in the [ImHex-Patterns](https://github.com/WerWolv/ImHex-Patterns) repo on your changes though. That usually gives a pretty decent indication if changes broke things or not

paxcut:
> I don't believe the original behaviour was correct.

I think it is clearly incorrect. An struct that has no template arguments should not be affected by what the parent has for template arguments. Issue #146 is just not making this point clear and it makes it look like the problem is one of priorities in evaluation. The evaluation of template arguments in the child should return empty because the child has no template arguments.
Consider this code
```cpp

#pragma example 01
import std.io;

struct Child {
    std::print("{}",size);
};

struct Parent1<auto size> {
     Child child;
};

struct Parent2 {
    u8 size;
    Child child;
};

Parent1<1> parent1@0;
Parent2 parent2@0;
```

It is clear that both parent1 and parent2 should return errors but only parent2 does currently. The changes proposed by this PR will not make parent1 return the same error parent2 returns so the real underlying problem will not be fixed.

PR #148, opened 2025-01-05
Error markers we displayed in the wrong place.

paxcut:
The problem existed for directives and the import statement. For example omitting the file to import or include resulted in an error marker located in the next line. The problem was that the tokens were being advanced prematurely due to a previous error fix that occurred when the same file was included twice. Imports suffered the same problem cause they were added after the bug fix. A pattern that included the same file twice was run, and it produced no errors. 

Other fixes and improvements include:
1) Moved the calculation of the longest line length of pattern source code into the lexer which already reads all the lines and can obtain it with little extra code. The code to use this improvement will be added in a soon to come pr for ImHex.

 2) The preprocessor still had code to remove tabs and extra newline characters that was both incorrect and is now unnecessary since the text editor is handling that directly. To test this code import the PE pattern which uses tabs instead of spaces

### DIFF
--- a/lib/source/pl/core/lexer.cpp
+++ b/lib/source/pl/core/lexer.cpp
@@ -134,12 +134,8 @@ namespace pl::core {
             result += character.value();
         }
 
-        if (m_sourceCode[m_cursor] == '\n') {
-            m_longestLineLength = compareCurrentWithLongest();
-            m_line++;
-            m_lineBegin = m_cursor;
+        if (hasTheLineEnded(m_sourceCode[m_cursor]))
             m_cursor++;
-        }
 
         return makeTokenAt(Literal::makeString(result), location, result.size());
     }
@@ -160,12 +156,8 @@ namespace pl::core {
             result += character.value();
         }
 
-        if (m_sourceCode[m_cursor] == '\n') {
-            m_longestLineLength = compareCurrentWithLongest();
-            m_line++;
-            m_lineBegin = m_cursor;
+        if (hasTheLineEnded(m_sourceCode[m_cursor]))
             m_cursor++;
-        }
 
         return makeTokenAt(Literal::makeString(result), location, result.size());
     }
@@ -325,12 +317,8 @@ namespace pl::core {
         }
         auto len = m_cursor - begin;
 
-        if (m_sourceCode[m_cursor] == '\n') {
-            m_longestLineLength = compareCurrentWithLongest();
-            m_line++;
-            m_lineBegin = m_cursor;
+        if (hasTheLineEnded(m_sourceCode[m_cursor]))
             m_cursor++;
-        }
 
         return makeTokenAt(Literal::makeComment(true, result), location, len);
     }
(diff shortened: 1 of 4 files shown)


lib/source/pl/core/lexer.cpp
```diff
@@ -135,6 +135,7 @@ namespace pl::core {
         }
 
         if (m_sourceCode[m_cursor] == '\n') {
+            m_longestLineLength = compareCurrentWithLongest();
```
WerWolv, line 138: This code repeats itself quite a lot now. Might make sense the move the whole thing into a function that just takes the current character as the input

lib/source/pl/core/preprocessor.cpp
```diff
@@ -235,21 +235,25 @@ namespace pl::core {
         // get include name
         auto *tokenLiteral = std::get_if<Token::Literal>(&m_token->value);
         if (tokenLiteral == nullptr || m_token->location.line != line ||  m_token->type != Token::Type::String) {
+            m_token--;
```
WerWolv, line 238: Can this m_token-- and m_token++ be moved into the `error` and `errorDesc` function? Would make more sense than having to manually remember to do it every time the function is used

paxcut: wouldn't that require that all the functions that use error will need to be changed as well?
Those functions are shared by the parser and the lexer and used quite a bit i think.
If you think it looks ugly I can look into ways to avoid having to move the current token back by not advancing it in the first place. I just thought this way was easier and less likely to create bugs if the tokens are not handled correctly.

lib/source/pl/core/lexer.cpp
paxcut: yes, possibly, originally the function there was actually just the code written out. I saw the repetition and made a function for it. I suppose I didn't go far enough.

lib/source/pl/core/preprocessor.cpp
WerWolv: Maybe a local wrapper would make sense then that handles the token iterator and calls the actual error and errorDesc function

lib/source/pl/core/lexer.cpp
paxcut: actually it doesn't quite repeat. there are a places  where
```cpp
                    if (peek(0) == '\n') {
                        m_longestLineLength = compareCurrentWithLongest();
                        m_line++;
                        m_lineBegin = m_cursor;
                        m_cursor++;
                        continue;
                    }
```
is used. in other places it is 
```cpp
        if (m_sourceCode[m_cursor] == '\n') {
            m_longestLineLength = compareCurrentWithLongest();
            m_line++;
            m_lineBegin = m_cursor;
            m_cursor++;
        }
```
 and in yet other places it is
```cpp
                   if(peek(0) == '\n') {
                m_longestLineLength = compareCurrentWithLongest();
                m_line++;
                m_lineBegin = m_cursor;
            }
```
moving just the comparison and the function call means we need to compare again to update the pointers. it needs to check every character of the file so it may not be a good idea to double the number of comparisons just to eliminate the function call duplication.

paxcut: ok, i can have the function return a bool that can be used to check the rest without duplication.

lib/source/pl/core/preprocessor.cpp
paxcut: Ok I wrote a function to handle the error calls when the token is decremented and then incremented and pushed the commit.Let me know if there are problems

lib/source/pl/core/lexer.cpp
paxcut: Wrote the code and pushed. Please review when you have some time.

WerWolv:
Thanks! Looks good now

PR #153, opened 2025-02-08
Fix conditional union size

zhabinski:
Union with conditional fields reports the size without including those fields.
See imhex issue for details https://github.com/WerWolv/ImHex/issues/1504


WerWolv:
Thank you!

PR #155, opened 2025-02-16
Fix error description vertical spacing

Naheulf:
Makes the error description closer to the main error message than to the next error.


Naheulf:
Or maybe the blank line should be used after every message?

WerWolv:
I think the way you proposed it now is good. That way everything that belongs together is displayed together

PR #163, opened 2025-05-09
Update libfmt to 11.2.0

henri-gasc:
As stated in the [libfmt release note](https://github.com/fmtlib/fmt/releases/tag/11.2.0): 
Deprecated `fmt::localtime` in favor of `std::localtime`.

This also fixes part of https://github.com/WerWolv/ImHex/issues/2225

### DIFF
--- a/lib/source/pl/lib/std/time.cpp
+++ b/lib/source/pl/lib/std/time.cpp
@@ -11,7 +11,7 @@
 
 namespace pl::lib::libstd::time {
 
-    static u128 packTMValue(std::tm tm) {
+    static u128 packTMValue(const std::tm &tm) {
         return
             (u128(tm.tm_sec)   << 0)  |
             (u128(tm.tm_min)   << 8)  |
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you! Is `fmt::gmtime` affected as well? Kinda weird that they'd only deprecate one and not the other

henri-gasc:
From what I can see, it is not impacted.
They probably deprecated `localtime` and not `gmtime`, because the latter is not thread-safe in std (as in, `This function may not be thread-safe.` as opposed to `This function may not be thread-safe. The [Microsoft CRT](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-localtime32-localtime64) implementation is thread-safe.`, but yeah, kinda weird

WerWolv:
Well we're using libstdc++, not MSVC so who knows if it is in our case. I just went ahead and replaced gmtime as well just for consistency. (Also fixed the build issue due to the different interface)

PR #164, opened 2025-05-12
Fix strict weak ordering

shewitt-au:
This is a fix for the ImHex bug "ImHex crashes when analysing any PE file #2221".  The fix requires changes to the Pattern Language and the ImHex UI. Revision would be wise. We want to avoid collateral damage. It's a big code base and I'm new to it and the compilers/build-systems used. And the bug is complex and low-level. I suspect this will fix other random crashes. The problem is caused by two issues:
 - The std::sort algorithm conjuring up garbage due to the sorting criteria not being a strict weak ordering. See  [this](https://github.com/Voultapher/sort-research-rs/blob/main/writeup/sort_safety/text.md) link. 
 - We sort shared_ptr&lt;ptrn::Pattern&gt; by pointer value, and the object is a clone.  In essence we're changing the values as we're sorting.

### DIFF
--- a/lib/source/pl/core/token.cpp
+++ b/lib/source/pl/core/token.cpp
@@ -112,6 +112,9 @@ namespace pl::core {
 
     std::strong_ordering Token::Literal::operator<=>(const Literal &other) const {
         return std::visit(wolv::util::overloaded {
+            [](std::shared_ptr<ptrn::Pattern> lhs, std::shared_ptr<ptrn::Pattern> rhs) {
+                return lhs->getOffset() <=> rhs->getOffset();
+            },
             []<typename T>(T lhs, T rhs) -> std::strong_ordering {
                 if (lhs == rhs) return std::strong_ordering::equal;
                 if (lhs < rhs) return std::strong_ordering::less;
@@ -158,7 +161,7 @@ namespace pl::core {
                 }
             },
             [](auto, auto) -> std::strong_ordering {
-                return std::strong_ordering::less;
+                return std::strong_ordering::equal;
             }
         }, *this, other);
     }


WerWolv:
Hey! Thanks a lot for looking into this. 
About your fix, will that actually lead to true strong ordering? Comparing the offsets of patterns with each other will lead to issues when two patterns are placed on the same address, which is easily possible. Would it make more sense to compare the pointer values with each other instead?

shewitt-au:
I was unsure how to sort the pattern vs pattern. I assumed offset was the offset of the structure in the binary. This is one of the acpects I'd like someone who knows what's going on to review.

shewitt-au:
You're better placed to make those choices. I've got not idea how this stuff works. It's a big code base. I've only scratched the surface. Impressice stuff. I'm low-level in esssence. This was some interesting debugging.

paxcut:
Maybe the value column should not be sorted at all. you cant compare pattern values in general so I can't imagine what a sorting should do.

shewitt-au:
But I guess the question is: what does a user expect when they soft by value?

shewitt-au:
Another option is lexicographically things as desired with the last comparison being the position of the rule in the pattern file.

paxcut:
a pattern value can be anything and much like apples and oranges they can't be compared. 

When considering what and how to sort,  we should be thinking about how the sorting  would be useful in practice. For example, sorting the starting addresses of the patterns  places them in the order they appear in the file. Similarly, sorting the size can help identify patterns in the data distribution. Sorting types allows one to compare similar results by placing them next to each other.

I am assuming that by to_string what is meant is  a lexicographic sorting of the resulting strings. It could be useful but not always.  If any form of sorting is to be implemented then I think that the only possible answer that is general enough would be some kind of user defined sorting function that accepts patterns does the rocket operator. That way if you wanted to sort lexicographically you could implement it that way and if you wanted to sort numerically you could too.

shewitt-au:
In reply to WerWolv's earlier question: "will that actually lead to true strong ordering"?
I think yes. If two offsets are the same they compare as std::strong_ordering::equal in Token::Literal::operator<=>. 
Down the stack in PatternDrawer::sortPatterns, this will converted to a false. If the parameters were reversed still a false, as required. So l<r would be false and r<l would be false. This is a strict weak ordering's equal.

lib/source/pl/core/token.cpp
```diff
@@ -158,7 +161,7 @@ namespace pl::core {
                 }
             },
             [](auto, auto) -> std::strong_ordering {
-                return std::strong_ordering::less;
```
shewitt-au, line 161: This is important. The only value can can be unconditionally returned from a strict weak compare function is false. The ImHex UI uses this function to implement its sorting function. Returning anything but std::strong_ordering::equal here will violate the strict weak ordering.

shewitt-au:
I have added two repos to my page:
  [WriteWatch](https://github.com/shewitt-au/WriteWatch)
  [MingW64-checked-pointer-vector](https://github.com/shewitt-au/MingW64-checked-pointer-vector)

These are the tools I wrote to verify the problem was not from external corruption.

PR #167, opened 2025-05-24
Cast to int format error v2

shewitt-au:
Another attempt at a fix for  [Cast to int format error (Format function no longer accepts enums) #166 ](https://github.com/WerWolv/PatternLanguage/pull/166)

Took me a few goes to get this. I the end it was pretty simple.

The usual disclaimers. I don't know the code base well etc.


WerWolv:
I actually wasn't aware of `std::enable_shared_from_this`! I believe that would actually solve our issue entirely. Let me play around with that a bit
Thanks a lot!

shewitt-au:
Passed test.

[PR](https://github.com/WerWolv/PatternLanguage/pull/169)

PR #169, opened 2025-05-25
cast-to-int-format-error-v3

shewitt-au:
Another attempt at a fix for  [Cast to int format error (Format function no longer accepts enums) #166 ](https://github.com/WerWolv/PatternLanguage/pull/166)

The previous attempt is [here](https://github.com/WerWolv/PatternLanguage/pull/167).

Both methods work, but this one is more in the spirit of how `shared_ptr`s are intended to be used and smells less of hack.

This one uses [enable_shared_from_this](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this).

### DIFF
--- a/lib/include/pl/patterns/pattern.hpp
+++ b/lib/include/pl/patterns/pattern.hpp
@@ -61,7 +61,7 @@ namespace pl::ptrn {
         friend class core::Evaluator;
     };
 
-    class Pattern {
+    class Pattern : public std::enable_shared_from_this<Pattern> {
     public:
         constexpr static u64 MainSectionId          = 0x0000'0000'0000'0000;
         constexpr static u64 HeapSectionId          = 0xFFFF'FFFF'FFFF'FFFF;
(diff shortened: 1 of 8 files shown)


WerWolv:
We'll need more changes than that sadly. Apparently this only works when called directly from another std::shared_ptr instance
image

shewitt-au:
Ah. That's what I was worried about when I implemented v2. Added v3 to demonstrate the option. Although I did get excited when, for me, it worked.

WerWolv:
Correction, I was using them wrong. The way you did it actually works!

shewitt-au:
Ah, that's a relief. Are all `Pattern`s owned by `shared_ptr`s? This is important. If UB will result, although from what I've seen you get `bad_weak_ptr`.

WerWolv:
Cloning currently returns a unique_ptr so there it wouldn't work. But I believe that's actually the only thing that does so we should be able to switch that over.
Other than that, Patterns are only produced by the evaluator runtime so that shouldn't be an issue

shewitt-au:
Assuming cloning is still needed all that's needed to fix that issue is for `clone` to use `shared_ptr` instead.

shewitt-au:
I'm not aware of this issue. So far I've been fixing problems that I encounter myself.

WerWolv:
Oh yeah the templates issue could be solved like this as well potentially. I'll look into it

shewitt-au:
No worries.

paxcut:
That would be awesome. being able to use recursion with templates makes some patterns so much easier to write. currently having to use parent and copying variables is doable but it can be hard to keep track of the information flow.

PR #170, opened 2025-05-31
Issue 2282

shewitt-au:
TBA

### DIFF
--- a/lib/include/pl/patterns/pattern_struct.hpp
+++ b/lib/include/pl/patterns/pattern_struct.hpp
@@ -15,14 +15,14 @@ namespace pl::ptrn {
             for (const auto &member : other.m_members) {
                 auto copy = member->clone();
 
-                copy->setParent(this);
+                copy->setParent(this->reference());
                 this->m_sortedMembers.push_back(copy.get());
                 this->m_members.push_back(std::move(copy));
             }
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternStruct(*this));
+            return std::shared_ptr<Pattern>(new PatternStruct(*this));
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> getEntry(size_t index) const override {
@@ -36,7 +36,7 @@ namespace pl::ptrn {
         void addEntry(const std::shared_ptr<Pattern> &entry) override {
             if (entry == nullptr) return;
 
-            entry->setParent(this);
+            entry->setParent(this->reference());
             this->m_sortedMembers.push_back(entry.get());
             this->m_members.push_back(entry);
         }
--- a/lib/include/pl/patterns/pattern.hpp
+++ b/lib/include/pl/patterns/pattern.hpp
@@ -282,9 +282,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] virtual core::Token::Literal getValue() const {
-            auto clone = this->clone();
-
-            return this->transformValue(std::move(clone));
+            return this->transformValue(const_cast<Pattern*>(this)->reference());
         }
 
         [[nodiscard]] virtual std::vector<std::pair<u64, Pattern*>> getChildren() {
@@ -537,15 +535,15 @@ namespace pl::ptrn {
             this->m_initialized = initialized;
         }
 
-        [[nodiscard]] const Pattern* getParent() const {
+        [[nodiscard]] std::shared_ptr<const Pattern> getParent() const {
             return m_parent;
         }
 
-        [[nodiscard]] Pattern* getParent() {
+        [[nodiscard]] std::shared_ptr<Pattern> getParent() {
             return m_parent;
         }
 
-        void setParent(Pattern *parent) {
+       void setParent(std::shared_ptr<Pattern> parent) {
             m_parent = parent;
         }
 
@@ -623,7 +621,7 @@ namespace pl::ptrn {
         core::Evaluator *m_evaluator;
 
         std::unique_ptr<std::map<std::string, std::vector<core::Token::Literal>>> m_attributes;
-        Pattern *m_parent = nullptr;
+        std::shared_ptr<Pattern> m_parent;
         u32 m_line = 0;
 
         std::set<std::string>::const_iterator m_variableName;
--- a/lib/include/pl/patterns/pattern_array_dynamic.hpp
+++ b/lib/include/pl/patterns/pattern_array_dynamic.hpp
@@ -20,7 +20,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternArrayDynamic(*this));
+            return std::shared_ptr<Pattern>(new PatternArrayDynamic(*this));
         }
 
         void setColor(u32 color) override {
@@ -133,7 +133,7 @@ namespace pl::ptrn {
 
             if (!entry->hasOverriddenColor())
                 entry->setBaseColor(this->getColor());
-            entry->setParent(this);
+            entry->setParent(this->reference());
 
             this->m_entries.emplace_back(entry);
         }
--- a/lib/include/pl/patterns/pattern_array_static.hpp
+++ b/lib/include/pl/patterns/pattern_array_static.hpp
@@ -16,7 +16,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternArrayStatic(*this));
+            return std::shared_ptr<Pattern>(new PatternArrayStatic(*this));
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> getEntry(size_t index) const override {
@@ -153,7 +153,7 @@ namespace pl::ptrn {
 
         void setEntries(std::shared_ptr<Pattern> &&templatePattern, size_t count) {
             this->m_template          = std::move(templatePattern);
-            this->m_template->setParent(this);
+            this->m_template->setParent(this->reference());
             this->m_highlightTemplates.push_back(this->m_template->clone());
             this->m_entryCount        = count;
 
--- a/lib/include/pl/patterns/pattern_bitfield.hpp
+++ b/lib/include/pl/patterns/pattern_bitfield.hpp
@@ -9,10 +9,15 @@ namespace pl::ptrn {
     public:
         using Pattern::Pattern;
 
+        /*virtual void setParent(Pattern *parent)
+        {
+            Pattern::setParent(parent);
+        }*/
+
         [[nodiscard]] const PatternBitfieldMember& getTopmostBitfield() const {
             const PatternBitfieldMember* topBitfield = this;
             while (auto parent = topBitfield->getParent()) {
-                auto parentBitfield = dynamic_cast<const PatternBitfieldMember*>(parent);
+                auto parentBitfield = dynamic_cast<const PatternBitfieldMember*>(parent.get());
                 if (parentBitfield == nullptr)
                     break;
 
@@ -53,7 +58,7 @@ namespace pl::ptrn {
 
     class PatternBitfieldField : public PatternBitfieldMember {
     public:
-        PatternBitfieldField(core::Evaluator *evaluator, u64 offset, u8 bitOffset, u8 bitSize, u32 line, PatternBitfieldMember *parentBitfield = nullptr)
+        PatternBitfieldField(core::Evaluator *evaluator, u64 offset, u8 bitOffset, u8 bitSize, u32 line, std::shared_ptr<PatternBitfieldMember> parentBitfield = nullptr)
                 : PatternBitfieldMember(evaluator, offset, (bitOffset + bitSize + 7) / 8, line), m_bitOffset(bitOffset % 8), m_bitSize(bitSize) {
             this->setParent(parentBitfield);
         }
@@ -65,7 +70,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternBitfieldField(*this));
+            return std::shared_ptr<Pattern>(new PatternBitfieldField(*this));
         }
 
         [[nodiscard]] u128 readValue() const {
@@ -162,7 +167,7 @@ namespace pl::ptrn {
         using PatternBitfieldField::PatternBitfieldField;
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternBitfieldFieldSigned(*this));
+            return std::shared_ptr<Pattern>(new PatternBitfieldFieldSigned(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
@@ -186,7 +191,7 @@ namespace pl::ptrn {
         using PatternBitfieldField::PatternBitfieldField;
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternBitfieldFieldBoolean(*this));
+            return std::shared_ptr<Pattern>(new PatternBitfieldFieldBoolean(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
@@ -245,7 +250,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternBitfieldFieldEnum(*this));
+            return std::shared_ptr<Pattern>(new PatternBitfieldFieldEnum(*this));
         }
 
         std::string formatDisplayValue() override {
@@ -282,7 +287,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternBitfieldArray(*this));
+            return std::shared_ptr<Pattern>(new PatternBitfieldArray(*this));
         }
 
         [[nodiscard]] u8 getBitOffset() const override {
@@ -422,7 +427,7 @@ namespace pl::ptrn {
                 if (!entry->hasOverriddenColor())
                     entry->setBaseColor(this->getColor());
 
-                entry->setParent(this);
+                entry->setParent(this->reference());
 
                 this->m_sortedEntries.push_back(entry.get());
             }
@@ -554,7 +559,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternBitfield(*this));
+            return std::shared_ptr<Pattern>(new PatternBitfield(*this));
         }
 
         [[nodiscard]] u8 getBitOffset() const override {
@@ -643,7 +648,7 @@ namespace pl::ptrn {
                 this->setBaseColor(this->m_fields.front()->getColor());
 
             for (const auto &field : this->m_fields) {
-                field->setParent(this);
+                field->setParent(this->reference());
                 this->m_sortedFields.push_back(field.get());
             }
         }
--- a/lib/include/pl/patterns/pattern_bitfield.hpp
+++ b/lib/include/pl/patterns/pattern_bitfield.hpp
@@ -9,11 +9,6 @@ namespace pl::ptrn {
     public:
         using Pattern::Pattern;
 
-        /*virtual void setParent(Pattern *parent)
-        {
-            Pattern::setParent(parent);
-        }*/
-
         [[nodiscard]] const PatternBitfieldMember& getTopmostBitfield() const {
             const PatternBitfieldMember* topBitfield = this;
             while (auto parent = topBitfield->getParent()) {
--- a/lib/include/pl/patterns/pattern_boolean.hpp
+++ b/lib/include/pl/patterns/pattern_boolean.hpp
@@ -10,7 +10,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, 1, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternBoolean(*this));
+            return std::shared_ptr<Pattern>(new PatternBoolean(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_enum.hpp
+++ b/lib/include/pl/patterns/pattern_enum.hpp
@@ -18,7 +18,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternEnum(*this));
+            return std::shared_ptr<Pattern>(new PatternEnum(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_error.hpp
+++ b/lib/include/pl/patterns/pattern_error.hpp
@@ -10,7 +10,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, size, line), m_errorMessage(std::move(errorMessage)) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternError(*this));
+            return std::shared_ptr<Pattern>(new PatternError(*this));
         }
 
         [[nodiscard]] std::string getFormattedName() const override {
--- a/lib/include/pl/patterns/pattern_float.hpp
+++ b/lib/include/pl/patterns/pattern_float.hpp
@@ -10,7 +10,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternFloat(*this));
+            return std::shared_ptr<Pattern>(new PatternFloat(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_padding.hpp
+++ b/lib/include/pl/patterns/pattern_padding.hpp
@@ -9,7 +9,7 @@ namespace pl::ptrn {
         PatternPadding(core::Evaluator *evaluator, u64 offset, size_t size, u32 line) : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternPadding(*this));
+            return std::shared_ptr<Pattern>(new PatternPadding(*this));
         }
 
         [[nodiscard]] std::string getFormattedName() const override {
--- a/lib/include/pl/patterns/pattern_pointer.hpp
+++ b/lib/include/pl/patterns/pattern_pointer.hpp
@@ -21,7 +21,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternPointer(*this));
+            return std::shared_ptr<Pattern>(new PatternPointer(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_signed.hpp
+++ b/lib/include/pl/patterns/pattern_signed.hpp
@@ -10,7 +10,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternSigned(*this));
+            return std::shared_ptr<Pattern>(new PatternSigned(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_string.hpp
+++ b/lib/include/pl/patterns/pattern_string.hpp
@@ -13,7 +13,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternString(*this));
+            return std::shared_ptr<Pattern>(new PatternString(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_union.hpp
+++ b/lib/include/pl/patterns/pattern_union.hpp
@@ -21,7 +21,7 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternUnion(*this));
+            return std::shared_ptr<Pattern>(new PatternUnion(*this));
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> getEntry(size_t index) const override {
@@ -35,7 +35,7 @@ namespace pl::ptrn {
         void addEntry(const std::shared_ptr<Pattern> &entry) override {
             if (entry == nullptr) return;
 
-            entry->setParent(this);
+            entry->setParent(this->reference());
             this->m_sortedMembers.push_back(entry.get());
             this->m_members.push_back(entry);
         }
--- a/lib/include/pl/patterns/pattern_unsigned.hpp
+++ b/lib/include/pl/patterns/pattern_unsigned.hpp
@@ -10,7 +10,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternUnsigned(*this));
+            return std::shared_ptr<Pattern>(new PatternUnsigned(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_wide_character.hpp
+++ b/lib/include/pl/patterns/pattern_wide_character.hpp
@@ -12,7 +12,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, 2, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternWideCharacter(*this));
+            return std::shared_ptr<Pattern>(new PatternWideCharacter(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
--- a/lib/include/pl/patterns/pattern_wide_string.hpp
+++ b/lib/include/pl/patterns/pattern_wide_string.hpp
@@ -13,7 +13,7 @@ namespace pl::ptrn {
             : Pattern(evaluator, offset, size, line) { }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternWideString(*this));
+            return std::shared_ptr<Pattern>(new PatternWideString(*this));
         }
 
         [[nodiscard]] core::Token::Literal getValue() const override {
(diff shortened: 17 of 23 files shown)


rsp4jack:
this is the backtrace:
```cpp
#0  0x00007ffff6cb1e9a in __cxxabiv1::__cxa_throw (obj=0x7fff84473b10, tinfo=0x7ffff6e86388 <typeinfo for std::bad_weak_ptr>, 
    dest=0x7ffff6ce4740 <std::bad_weak_ptr::~bad_weak_ptr()>) at /usr/src/debug/gcc/gcc/libstdc++-v3/libsupc++/eh_throw.cc:81
#1  0x00007ffff7ae4a46 in __wrap___cxa_throw (thrownException=0x7fff84473b10, type=0x7ffff6e86388 <typeinfo for std::bad_weak_ptr>, 
    destructor=0x7ffff6ce4740 <std::bad_weak_ptr::~bad_weak_ptr()>) at /home/rsp4jack/workspace/ImHex/lib/trace/source/exceptions.cpp:25
#2  0x00007ffff738e6fb in std::__throw_bad_weak_ptr () at /usr/include/c++/15.1.1/bits/shared_ptr_base.h:95
#3  0x00007ffff74a1250 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count (this=<optimized out>, __r=...)
    at /usr/include/c++/15.1.1/bits/shared_ptr_base.h:1246
#4  std::__shared_ptr<pl::ptrn::Pattern, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<pl::ptrn::Pattern, void> (this=<optimized out>, __r=...)
    at /usr/include/c++/15.1.1/bits/shared_ptr_base.h:1555
#5  std::shared_ptr<pl::ptrn::Pattern>::shared_ptr<pl::ptrn::Pattern, void> (this=<optimized out>, __r=std::weak_ptr<pl::ptrn::Pattern> (empty) = {...})
    at /usr/include/c++/15.1.1/bits/shared_ptr.h:380
#6  std::enable_shared_from_this<pl::ptrn::Pattern>::shared_from_this (this=0x7fff8446fdd8) at /usr/include/c++/15.1.1/bits/shared_ptr.h:934
#7  pl::ptrn::Pattern::reference (this=0x7fff8446fdd0) at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/include/pl/patterns/pattern.hpp:119
#8  pl::ptrn::PatternStruct::PatternStruct (this=0x7fff8446fdd0, other=...)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/include/pl/patterns/pattern_struct.hpp:18
#9  0x00007ffff74a12e8 in pl::ptrn::PatternStruct::clone (this=0x7fff844737a0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/include/pl/patterns/pattern_struct.hpp:25
#10 0x00007ffff7424da9 in pl::core::ast::ASTNodeArrayVariableDecl::createStaticArray (this=this@entry=0x7fff840f3020, evaluator=evaluator@entry=0x5555568a38b0, 
    outputPattern=std::shared_ptr<pl::ptrn::Pattern> (empty) = {...})
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_array_variable_decl.cpp:228
#11 0x00007ffff7425305 in pl::core::ast::ASTNodeArrayVariableDecl::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0, 
    resultPatterns=std::vector of length 1, capacity 1 = {...})
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_array_variable_decl.cpp:91
#12 0x00007ffff7481a87 in pl::core::ast::ASTNodeConditionalStatement::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_conditional_statement.cpp:31
#13 0x00007ffff7481a87 in pl::core::ast::ASTNodeConditionalStatement::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_conditional_statement.cpp:31
#14 0x00007ffff7481a87 in pl::core::ast::ASTNodeConditionalStatement::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_conditional_statement.cpp:31
#15 0x00007ffff7481a87 in pl::core::ast::ASTNodeConditionalStatement::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_conditional_statement.cpp:31
#16 0x00007ffff7481a87 in pl::core::ast::ASTNodeConditionalStatement::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_conditional_statement.cpp:31
#17 0x00007ffff7481a87 in pl::core::ast::ASTNodeConditionalStatement::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_conditional_statement.cpp:31
#18 0x00007ffff7481a87 in pl::core::ast::ASTNodeConditionalStatement::createPatterns (this=<optimized out>, evaluator=0x5555568a38b0)
    at /home/rsp4jack/workspace/ImHex/lib/external/pattern_language/lib/source/pl/core/ast/ast_node_conditional_statement.cpp:31

```

paxcut:
One simple way to create the problem
```cpp
#pragma example 00 00 00 00
import std.mem;

std::mem::Bytes<4> b@0;
```
The pattern will run but the value shown in the pattern data window is bad_weak_ptr

image

lib/include/pl/patterns/pattern_struct.hpp
```diff
@@ -15,14 +15,14 @@ namespace pl::ptrn {
...
+                copy->setParent(this->reference());
                 this->m_sortedMembers.push_back(copy.get());
                 this->m_members.push_back(std::move(copy));
             }
         }
 
         [[nodiscard]] std::shared_ptr<Pattern> clone() const override {
-            return std::unique_ptr<Pattern>(new PatternStruct(*this));
+            return std::shared_ptr<Pattern>(new PatternStruct(*this));
         }
```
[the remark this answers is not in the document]
paxcut: shouldn't clone() just be returning shared_from_this like reference?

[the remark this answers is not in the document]
rsp4jack: @paxcut IMO clone() should return a new instance.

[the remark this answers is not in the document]
rsp4jack: A possibly solution is initalize the shared_ptr first with a default constructed instance, then copy assign *this to it.

shewitt-au:
This is a bind... I'll have to think a little on this.

lib/include/pl/patterns/pattern_struct.hpp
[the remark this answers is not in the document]
paxcut: shouldnt there be a call to make_shared somewhere?

shewitt-au:
This is a serious issue. Basically we cant call `std::shared_from_this` from constructors (or any methods call from them). When I introduced `std::shared_from_this` [here](https://github.com/WerWolv/PatternLanguage/pull/167) I was used a different method which was my preference. Perhaps it should be revisited..

paxcut:
why cant you call make_shared() to create the initial shared pointer?

shewitt-au:
As I understand it, even if we used `std::make_shared' 'std::shared_from_this' still requires a constructed object.  `std::make_shared' is more efficient as the control block and the object are enplaced in a single allocation. And it's more exception safe.

rsp4jack:
@shewitt-au How about this? It should avoid the problem. There may not be a suitable constructor or copy assignment, if so, just create one.

```cpp
[[nodiscard]] std::shared_ptr<Pattern> clone() const override {
    auto result = std::make_shared<PatternStruct>(<something>);
    *result = *this;
    return std::static_pointer_cast<Pattern>(result);
}
```

paxcut:
constructors cannot call shared_from_this()

rsp4jack:
> if <something>'s constructor uses std::shared_from_this

Actually I want to construct an instance using a constructor which does nothing. If there isn't one, Just create a constructor which does nothing. This is a member function, so the new constructor could be private.

shewitt-au:
@rsp4jack Yeah, something like that would seem to be the way to go. Make the constructors trivial and private and move construction to factory methods.

paxcut:
From what I read it seems that in order to use shared_from_this the object in question must already have been used to create a shared pointer using any of the methods available to do so. Once this initial pointer exists, further instances can be obtained using shared_from_this().

It also seems to me that we could set a member flag that indicates if a shared pointer exists for the object in question initially set to false. Before calling shared_from_this() the flag is checked and if false then the shared pointer object is created and the flag is set to true. This way the pointers are only created when needed. I am probably over simplifying things. this is my first experience with std shared from this

rsp4jack:
I just think up a scenario (seems rare):

```cpp
auto* m_pattern = new PatternFooBar{*pattern_with_many_children};
// now m_pattern's children have shared_ptr pointing to m_pattern
```

and...

```cpp
// kill all the children!
kill_children(m_pattern);
```

Then `m_pattern` is gone! (because children are the only owners of m_pattern)

So, since children should not exist if their parent isn't exist, why a pattern must hold a shared_ptr pointing to its parent? A weak_ptr is enough (or even simply a bare pointer). Maybe there are some problems behind the scene, where a pattern dies with its children still alive.

rsp4jack:
And there is a circular reference problem.

A pattern hold a shared_ptr to its parent, while its parent hold shared_ptrs to its children: 

https://github.com/WerWolv/PatternLanguage/blob/db3c9475149edf93031e33a0b959e2087ca98092/lib/include/pl/patterns/pattern_struct.hpp#L230

The parent's lifetime should be longer than its children, so the pointer to parent should be a weak_ptr or bare pointer. Using a shared_ptr is a bad practice.

lib/include/pl/patterns/pattern_struct.hpp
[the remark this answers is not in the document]
shewitt-au: I think I have fixed the issue. There may be others but the framework is in-place to fix bugs as they are found fairly easy. Uses some funky template code, so testing would be wise.

shewitt-au:
Yeah. The problem I'm looking at now is that `rererence()` returns a `std::shared_ptr<Pattern>` but we need a `std::shared_ptr<PatternBitfieldField>`. I could put a cast in but I'm trying to find a better solution.

The fundamental problem is the mix between using an object managed be `std::shared_ptr`s, `std::unique_ptr`s and raw pointers. Using reference counted pointers fixes a lot of bugs, but it's causing problems in the tests.

Re the tests: I can't find that target:
```
steve@SteveWin MINGW64 /c/projects/SteveImHex/lib/external/pattern_language
$ cmake --build build-release --target help | grep test
external/libwolv/libs/testing/edit_cache: phony
external/libwolv/libs/testing/rebuild_cache: phony
external/libwolv/libs/testing/list_install_components: phony
external/libwolv/libs/testing/install: phony
external/libwolv/libs/testing/install/local: phony
external/libwolv/libs/testing/install/strip: phony
liblibwolv-testing.a: phony
libwolv-containers-test: phony
libwolv-containers-test.exe: phony
libwolv-hash-test: phony
libwolv-hash-test.exe: phony
libwolv-io-test: phony
libwolv-io-test.exe: phony
libwolv-math_eval-test: phony
libwolv-math_eval-test.exe: phony
libwolv-net-test: phony
libwolv-net-test.exe: phony
libwolv-testing: phony
libwolv-types-test: phony
libwolv-types-test.exe: phony
libwolv-utils-test: phony
libwolv-utils-test.exe: phony
```

paxcut:
You seem to be under the impression that the problem is caused by constructors calling shared_from_this. The only constructors that can call it are copy constructors and that is taken care of by the copy constructor of Pattern initializing using enable_shared_from_this(other). Under the debugger, all the copy constructors run fine, the problem arises when the m_pattern variable is assigned a value without setting shared pointer to manage the pattern being used as parent. later on that will trigger the exception.

paxcut:
m_parent does need to be made a weak pointer if only because we use shared_from_this but that is easily handled by changing the get_parent and set_parent functions to the appropriate types like this which i think is the same code you have
```cpp
 [[nodiscard]] const std::shared_ptr<Pattern> getParent() const {
            return m_parent.lock();
        }

        [[nodiscard]] std::shared_ptr<Pattern> getParent() {
            return m_parent.lock();
        }

        void setParent(std::shared_ptr<Pattern> parent) {
            m_parent = parent;
        }
     ```

**Comment by shewitt-au** [comment]
It's likely it need the be weak or raw. Obviously the code you posted shows that a child references it's parent. It's hard to imaging a scenario when a parent doesn't reference it's children. This being the case, and if we're going to use reference counting for lifetime managment, we've got a cycle. Parent keeps child alive; child keeps parent alive. A memory leak.

**Comment by shewitt-au** [comment]
It is enough to create a cycle. Timing doesn't matter, only that once the cycle is established the objects in the cycle, unless an external event intervenes to break the cycle; all of the objects in the cycle are immortal.
In the diagrams below of an arrow points to an object (references or points to it) it will not be deleted.
A -> B
B will live as long as A does.
A <- B
A will live as long as B does
A<->B
They both point to each other. No external reference is required to keep them alive. Thay are immortal.

**Comment by shewitt-au** [comment]
"All the talk about cycles started when changes on your pr turned m_parent from a raw pointer to a shared_pointer."

It's not a `std::shared_ptr`.

```c++
    private:
        friend pl::core::Evaluator;

        core::Evaluator *m_evaluator;

        std::unique_ptr<std::map<std::string, std::vector<core::Token::Literal>>> m_attributes;
        std::weak_ptr<Pattern> m_parent; // <------ Here
        u32 m_line = 0;
```

paxcut:
no problem, it is hard to convey emotions over written text but i assure you that i dont consider the exchange to be anything but anexchange of ideas and points of views.  In my opinion  we are not arguing about anything worth arguing about. There are things we should be talking about instead though.

shewitt-au:
Ok. Agreed.

shewitt-au:
It was my intention the get all of them. Sounds like I make have missed some. The code in "construct_shared_object.hpp" supports making the constructors non-public (via the the `BEFRIEND_SHARED_OBJECT_CREATOR ` macro), which would make such errors trigger a compiler error.

paxcut:
Clone is not always called from constructors. in fact it is safe to call from constructors if they are copy constructors. look in both builtin_types.cpp files from the imhex (client) changes of your pr, each one defines a clone function that uses make_shared

paxcut:
the other one is in dissasembler. They dont need to be fixed because they work just fine. In fact all clones work fine using make_shared

shewitt-au:
Ok. This PR is now close to where I want it to be for now.

The two-stage construction may be contentious, and I get that, but it was used in four pattens to fix (hopefully) real bugs. 

`<pl/helpers/construct_shared_object.hpp>` is not intended to be an end onto itself. But, for now at least, it has served its purpose.

Firstly, it is a debugging tool. It has made it easier for me to update `Pattern` code which was causing bugs due to using `shared_from_this` in constructors. Assuming all the creation code has been updated to use `construct_shared_object` I can make the construction two-stage by simply adding a `post_construct` method.

It allowed me to find `Pattern` creation that bypasses `construct_shared_object` by making the `Pattern`’s constructions `protected` and using the `BEFRIEND_SHARED_OBJECT_CREATOR` macro to allow `construct_shared_object` to still create it. This will not pick up such bypasses in a Pattern class’s code itself however.

I think perhaps we need think using the factory patten for Patten creation in future.
But, that said, I’ve not has any crashes in the while importing patterns and associated activity.

shewitt-au:
Ok. I added to to check for zombie `Pattern`'s on app exit. There's nothing worse than fixing deleted object references by creating zombies. I did not find any. Looks the the `Pattern` lifetime management is sound.

I also changes a static cast to a dynamic cast in the tests.

I think I'm done with this PR.

shewitt-au:
OK. I'll take this bit by bit because your questions, justifiably, require answers. Sometimes long ones.

> If two step initialization is being contested but you claim it is necessary then you should be stating what examples need it in order to run so that they can be tested without the two steps in order to validate your claim.

This depends on what you mean by two-stage construction, I guess. I'll come back to this later.

The crux of the issue is we can't use `shared_from_this` from constructors, and currently we do. On that, I've known of `std::enable_shared_from_this` since it was only in the Boost libraries, but it's only since this issue that I more fully appreciate its limitations. 

All the places I've used two-stage creation can be found by searching in my PR for `post_construct`:
- lib\include\pl\patterns\pattern_array_dynamic.hpp
- lib\include\pl\patterns\pattern_array_static.hpp
- lib\include\pl\patterns\pattern_bitfield.hpp
- lib\include\pl\patterns\pattern_struct.hpp

The first issue is that `clone` can call `shared_from_this`, and we sometimes clone from constructors. This stack trace is from my PR code. Anything in  `post_construct` was copied from the constructor. I have annotated the  stack trace with code snippets.
```
Pattern::reference(pl::ptrn::Pattern * const this) (c:\projects\SteveImHex\lib\external\pattern_language\lib\include\pl\patterns\pattern.hpp:125)
{
	std::shared_ptr<Pattern> reference() { return shared_from_this(); }
}

libimhex.dll!pl::ptrn::PatternStruct::post_construct(pl::ptrn::PatternStruct * const this, const pl::ptrn::PatternStruct & other) (c:\projects\SteveImHex\lib\external\pattern_language\lib\include\pl\patterns\pattern_struct.hpp:21)
{
	void post_construct(const PatternStruct &other) {
            for (const auto &member : other.m_members) {
                auto copy = member->clone();

                copy->setParent(this->reference()); // <------ Here
                this->m_sortedMembers.push_back(copy.get());
                this->m_members.push_back(std::move(copy));
            }
        }
}
 .
 .
 .
libimhex.dll!pl::ptrn::PatternStruct::clone(const pl::ptrn::PatternStruct * const this) (c:\projects\SteveImHex\lib\external\pattern_language\lib\include\pl\patterns\pattern_struct.hpp:28)
{
	[nodiscard]] std::shared_ptr<Pattern> clone() const override {
            return construct_shared_object<PatternStruct>(*this);
        }
}

```

The second is `setParent' (also in the above example).

I only added `post_construct`s in resopnse to actual crashes that I reproduced locally. Given this it has taken some time for me to rediscover them. Sorry about that.

Back to the "I'll come back to this late":
 Say currently, like in the above code, `setParent` is called in a constructor. In this case settiing the parent is part of object construction. Now say the code is refactored so that `setParent` is called after construction; now setting the parent is not part of object construction.

> you claim it is necessary

`post_construct` is only intended to be a tool to help address these issues quickly, and possibly as a transitional measure to stabalise the code until it is no longer needed. I make no claims beyond that.

paxcut:
Obviously using a public function from a friend class is the same as using a public function from the class itself. You object_creator is now a factory to all the classes derived from Pattern.

paxcut:
> Generally, yes. But not when constructors are non-public (which is not that unusual). Then it's the only choice and definitially within guidelines.

That is incorrect. As I stated there are ways to call make_share when the constructors are not public. Guidelines for c++ is to always use make_shared when creating shared pointers. A simple google search will let you find ways to do so.

> I'd be interested in seeing your code. You have me at a disadvantage here. I’d like to see your code.

It is not my code, it is your code with all the parts that are not necessary removed.

shewitt-au:
>Obviously using a public function from a friend class is the same as using a public function from the class itself. You object_creator is now a factory to all the classes derived from Pattern.

I didn't mean to be condescending. From the wording of your question, it was not obvious that you were considering this.

shewitt-au:
>That is incorrect. As I stated there are ways to call make_share when the constructors are not public. Guidelines for c++ is to always use make_shared when creating shared pointers. A simple google search will let you find ways to do so.

I don't know of any. If you do, please share.

paxcut:
> I didn't mean to be condescending. From the wording of your question, it was not obvious that you were considering this.

My question was that I was confused because you were saying that we should add factory functions when you already had added them by making constructors non-public. I was obviously aware that create_object is a class and i didn't think it was necessary to go through the detail explaining that but apparently I did. Sorry for that.

shewitt-au:
No worries man.

paxcut:
> If you code is better I'd like to see it.

it isn't my code, I didn't create it. you did and I only removed the parts that were not needed. The code is not better or worse. the number of changes is smaller and that will help you get the changes merged faster. I have no interest in fixing this bug. I don't really care if you  keep the code as it is. If you get it merged fast then all is good. I have rewritten prs that were large after waiting a long time for it to be merged to introduce changes a bit at a time, so this is just my experience and it may not apply to you. I am not trying to prevent your code from being added, quite the contrary. I wont be posting code to compete against yours either. this is not about that and i thought you understood this. I guess I need to work on my ability to explain myself but im a bit too old for that.

shewitt-au:
I have to sleep now, but I gave it a quick read. It does indeed look promising. I'll give it more thorough read tomorrow. Thanks for the link.

shewitt-au:
>I wont be posting code to compete against yours either. this is not about that and i thought you understood this.

I didn't interpret this a competition. I'm not sure why imply I do, assuming I read the non too subtle subtext correctly. If not, my appologies.

shewitt-au:
I'm ending this. Just becaue you looked and found it easy doesn't negate the fact that I looked and found it hard. I even used ChatGPT. I looked, and I looked hard.  Your "search" link does is not even a link to a search, rather the end result of search(s).

shewitt-au:
I appoligise for that. Clicking on the link tool me directly to the page. That said, the rest of my comments I stand by.

rsp4jack (approved):

rsp4jack:
I just run the initial pattern (where I found the problem and created https://github.com/WerWolv/ImHex/issues/2282) and there is no problem.

Current code looks well. The `construct_shared_object` and `post_construct` is good, I think this is probably the best solution for now to fix this bug. 

IMO This PR is almost done. Thanks for your effort in it.

shewitt-au:
@paxcut 

Re the [article](https://embeddedartistry.com/blog/2017/01/11/stdshared_ptr-and-shared_from_this/) you provided: 

```c++
template<typename... T>
static std::shared_ptr<SharedThing> create(T&&... t)
{
    // C++ core guidelines say we shouldn't do this:
    //return std::shared_ptr<SharedThing>(new SharedThing(std::forward<T>(t)...));

    // But this doesn't work because our constructor is private, and we want that
    // for use with enable_shared_from_this
    //return std::make_shared<SharedThing>(std::forward<T>(t)...);

    // So we'll make a "fake" derived class that constructs our shared thing
    // using a public constructor...
    struct EnableMakeShared : public SharedThing {
        EnableMakeShared(T&&... arg) : SharedThing(std::forward<T>(arg)...) {}
    };

    // and use that type with make_shared
    return std::make_shared<EnableMakeShared>(std::forward<T>(t)...);
}
```

One initial problem occurs to me (perhaps it's a solvable one). ImHex uses `dynamic_cast` to determine a `Pattern`s type at runtime. The approach above messes up the class hierarchy.

Antother one, which I don't think would be of real concern, is that it would not work on `final` classes. Although, like I've done in my code as it stands, `template`s could probably we used to select an alternate inpementation in this case.

shewitt-au:
@paxcut 

Re the [article](https://embeddedartistry.com/blog/2017/01/11/stdshared_ptr-and-shared_from_this/) you provided: 

```c++
template<typename... T>
static std::shared_ptr<SharedThing> create(T&&... t)
{
    // C++ core guidelines say we shouldn't do this:
    //return std::shared_ptr<SharedThing>(new SharedThing(std::forward<T>(t)...));

    // But this doesn't work because our constructor is private, and we want that
    // for use with enable_shared_from_this
    //return std::make_shared<SharedThing>(std::forward<T>(t)...);

    // So we'll make a "fake" derived class that constructs our shared thing
    // using a public constructor...
    struct EnableMakeShared : public SharedThing {
        EnableMakeShared(T&&... arg) : SharedThing(std::forward<T>(arg)...) {}
    };

    // and use that type with make_shared
    return std::make_shared<EnableMakeShared>(std::forward<T>(t)...);
}
```

One initial problem occurs to me. The class will now be a `EnableMakeShared`. Would this make debugging confusing for example? I'm not sure.

Anohter one, which I don't think would be of real concern, is that it would not work on `final` classes. Although, like I've done in my code as it stands, `template`s could probably we used to select an alternate impementation in this case.

shewitt-au:
I think I've found an even better solution while reseaching my concerns above. It can be found [here](https://stackoverflow.com/questions/8147027/how-do-i-call-stdmake-shared-on-a-class-with-only-protected-or-private-const).  The chain of authors going back some time. The version linked seems sound and would result in a single allocation for the object and the control block.

```c++
class safe_enable_shared_from_this : 
    public std::enable_shared_from_this<safe_enable_shared_from_this> 
{
protected:
    safe_enable_shared_from_this() noexcept = default;
    safe_enable_shared_from_this(safe_enable_shared_from_this&&) noexcept = default;
    safe_enable_shared_from_this(const safe_enable_shared_from_this&) noexcept = delete;
    safe_enable_shared_from_this& operator=(safe_enable_shared_from_this&&) noexcept = default;
    safe_enable_shared_from_this& operator=(const safe_enable_shared_from_this&) noexcept = delete;

public:
    virtual ~safe_enable_shared_from_this() noexcept = default;

protected:
    template <typename T>
    struct Allocator : public std::allocator<T>
    {  
        template<typename TParent, typename... TArgs>
        void construct(TParent* parent, TArgs&&... args) { 
            ::new((void *)parent) TParent(std::forward<TArgs>(args)...);
        }
    };
    
    template <typename T, typename... TArgs>
    static inline auto create(TArgs&&... args) -> ::std::shared_ptr<T> 
    {
        return std::allocate_shared<T>(Allocator<T>{}, std::forward<TArgs>(args)...);
    }

public:
    template <typename TSelf>
    auto inline shared_from_this(this TSelf&& self) noexcept
    {
        return std::static_pointer_cast<std::remove_reference_t<TSelf>>(
            std::forward<TSelf>(self).std::template enable_shared_from_this<safe_enable_shared_from_this>::shared_from_this());
    }
    
    template <typename TSelf>
    auto inline weak_from_this(this TSelf&& self) noexcept -> ::std::weak_ptr<std::remove_reference_t<TSelf>>
    {
        return std::static_pointer_cast<std::remove_reference_t<TSelf>>(
            std::forward<TSelf>(self).std::template enable_shared_from_this<safe_enable_shared_from_this>::weak_from_this().lock());
    }
};
```
I'm going to code it up and see how it goes.

Here's the code

paxcut:
I went back to master code and looked into why the errors occur. There are four examples that generate errors. 
1) static array of bitfields, 
2) json variable used as reference and 
3) custom inspector rows 
4) exporting patterns. 

There are four classes derived from Pattern that have non-trivial copy constructors. Although there are calls to shared_from_this() inside them, they are never the ones that belong to the this pointer. They are called for members, for parents, for array elements, etc.. and they are always called post construction.
So while it is constructing the pattern for an array, the patterns for all the elements of the array were already constructed previously and it is safe to call shared_from_this for them. There isn't a single case of shared_from_this being called from its own constructor. The major problem with the way the copy constructors are setup currently is that it is very easy to go into infinite loops that overflows the stack. 

One important fact is that all the copy constructors are called from clone and only from clone, so instead of creating new functions, I have moved all the code that was on the copy constructors onto clone to avoid infinite loops.

All the errors in master occur because either unique pointers are being used instead of shared or because m_parent was assigned a raw pointer and never turned into a shared pointer before shared_from_this is called or because of infinite loops.

Starting from master i added code to reference() that throws an error if the weak pointer is expired. that finds all the cases when shared pointers need to be created. 
I have fixed all the cases related to the four examples above and all the code runs just as before except for maybe the json example where the reference use zeroes it out which occurs in your version as well. I need to investigate further on that. 

Sorry about the long post. hopefully it will be my last here.

shewitt-au:
@paxcut I had a look at your changes.

```c++
std::shared_ptr<Pattern> reference() {
    auto weakPtr = weak_from_this();
        if (weakPtr.expired()) {
            core::err::E0001.throwError("Cannot call shared_from_this if this is not shared.");
        }
           
    return shared_from_this();
}
````
This is just obviously a good idea.

Your code certainly has the advantage of simplicity. There's no extra code like `construct_shared_object.hpp` in my version.

Many of the fixes are like my code, but instead of moving the code into `post_construct` it's been moved to `clone`. You've used two-stage construction implictly instead of explictly, from the perspective of how the code was before the changes. My PR has the advantage of the intent being explicit. If you see a `post_construct` method you know that the code in it, in a more perfect world, would be in a constructor (and in fact used to be), but can't be because of the limitations of `std::enable_shared_from_this`. Although I don't want to overstate this: it is fairly obvious that a `clone` method constructs an object. 

There is nothing to stop misuse however. If copy constructors are only used from `clone` it could be made less error prone by making the copy constructor `protected`, but then `std::make_shared` could not be used. That said, If someone copy constructs an object without using `clone` they no longer get a fully constructed object, much of the the code has been removed from the constructor and into `clone`. This the case with my code too (although all my constructors are `protected`, so the scope of code is exposed code is reduced), but the presence of a `post_construct` method makes this explicit.

I have updated my code in response to our conversations. Although we may have butted heads at times, My PR, regardless whether it is adopted or not, is better as a result.

All the constructors are `protected`; abuse can only happen from within. The code in `construct_shared_object.hpp` has been updated so that, even though the constructors are non-public, only a single allocation is used (like with `std::make_shared` with public constructors).  When `shared_from_this` and `weak_from_this` methods are called from derived classes the `std::shared_ptr` return now wraps the derived type.

I understand the simplicity argument. And it has merit. My perspective is that if mistakes are harder to make, less mistakes will be made.

**Comment by shewitt-au** [comment]
I understand. Although personally I feel my changes are safer than the alternative, I don't blame you for being skeptical.

File: lib/include/pl/helpers/construct_shared_object.hpp [hunk 2]
```diff
@@ -26,61 +25,106 @@ construction on demand.
...
-
-    // The actual creation functions.
-
+// safe_enable_shared_from_this was found here:
+//  https://stackoverflow.com/questions/8147027/how-do-i-call-stdmake-shared-on-a-class-with-only-protected-or-private-const
+// 
+// Posted by stackoverflow member Carsten.
+// He adapted old code from stackoverflow member Zsolt Rizsányi,
+// who in turn says it was invented by Jonathan Wakely (GCC developer).
+class safe_enable_shared_from_this : public std::enable_shared_from_this<safe_enable_shared_from_this> {
```
rsp4jack, line 57: I think the name is kinda confusing: "a safe enable_shared_from_this"

It not only is another enable_shared_from_this without CRTP, but also enables `BEFRIEND_CONSTRUCT_SHARED_OBJECT`. IMO the class's job, besides enabling shared_from_this, is only hiding Allocator from public access.

And what does "safe" mean?

I think it is needed to have a better name for this class.

shewitt-au: Yeah. Those are fair points. To be honest the name as bother me a bit too. As I cited in the code, I did notwrite `safe_enable_shared_from_this`. I think the "safe" refers to the abiliy to make the constructors non-public. I'll try to come up with a better name. Any suggestions?

shewitt-au:
@paxcut

>Let me be 100% clear about this. Your changes are obviously safer than mine or the code that was implemented before. It isn't even debatable that it is obvious. That being said, I fail to see how that has any relevance on what needs to be done though. The bug needs to be fixed now but the code doesn't need to be made safer now. It can and it will be made safer later.

I already said I understood your skepticism. And if you want to get a fix in as a matter of urgency (which feels safer) I understand that too.  But since we're talking about it I'll elaborate on the reasons I went to the effort to make the code safer. To make the kind of mistakes that caused these bugs harder to make in future.

>The total entropy of an isolated system always increases

 — 2nd law of thermodynamics

It takes effort to keep anything ordered, software being no exception. It's easy to make rapid progress at the expense of long term progress and maintainability. With your changes `PatternArrayDynamic`, `PatternArrayStatic` and `PatternStruct` (from memory) now have (publically accessbile) broken copy constructors. Too me it feels like fixing the bug, not the code

>Everything should be made as simple as possible, but not simpler

— Einstein

OK. I've satiated the urge to quote. Like I said earlier, I understand that in software development sometimes you need to get fixes out quickly and that may involve suboptimal code.

lib/include/pl/helpers/construct_shared_object.hpp
```diff
@@ -26,61 +25,106 @@ construction on demand.
...
+	{
+		return std::static_pointer_cast<std::remove_reference_t<TSelf>>(
+			std::forward<TSelf>(self).std::template enable_shared_from_this<safe_enable_shared_from_this>::shared_from_this());
+	}
+	
+	template <typename TSelf>
+	auto inline weak_from_this(this TSelf&& self) noexcept -> ::std::weak_ptr<std::remove_reference_t<TSelf>>
+	{
+		return std::static_pointer_cast<std::remove_reference_t<TSelf>>(
+			std::forward<TSelf>(self).std::template enable_shared_from_this<safe_enable_shared_from_this>::weak_from_this().lock());
```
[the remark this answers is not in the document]
shewitt-au: Bacause `std::static_pointer_cast` does not work on `std::weak_ptr`'s.

lib/include/pl/helpers/construct_shared_object.hpp
rsp4jack: How about `shared_constructible_enable_shared_from_this`? I know it is long, but a complex thing deserves a long name.

lib/include/pl/helpers/construct_shared_object.hpp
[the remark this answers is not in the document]
shewitt-au: The overhead in minimal since no allocation is taking place.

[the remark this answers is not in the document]
shewitt-au: I use LLMs a bit these days. Often I regard them as just better search engines. Sometimes they do provide valuable insights staingt out if the gate. Then there are times when they help, but you have to beat them into submission. And then the times they just spout garbage.

lib/include/pl/helpers/construct_shared_object.hpp
shewitt-au: The "shared" being used twice is perhaps non needed. How about the slightly shorter `constructible_enable_shared_from_this`?

rsp4jack: Looks kinda strange. `enable_shared_from_this` is a whole one, and has nothing to do with `shared_constructible`.

rsp4jack: > There are only two hard things in Computer Science: cache invalidation and **naming things**.
>
> -- Phil Karlton

shewitt-au: I asked ChatGPT. Some of it's suggestions:

- `enable_shared_from_this_safe`
- `shared_from_this_factory` 
- `constructible_shared_from_this` 
- `SharedObjectBase`

shewitt-au: How about `enable_shared_from_this_factory'?

paxcut: Maybe make_shared?

shewitt-au: That could be confusing given `std::make_shared`.

paxcut:
> I understand that in software development sometimes you need to get fixes out quickly and that may involve suboptimal code. I accept that pragmatism is a virtue.

As long as you understand I won't feel the need to explain things again and I hope that you don't feel the need of  having to explain things to ma that I already know either. If you do, I am afraid I will have to explain myself again despite your claims of understanding.

lib/include/pl/helpers/construct_shared_object.hpp
paxcut: make_shared_from_private_constructor?

shewitt-au: `enable_shared_from_private_constructor`. Yeah, I don't mind that one.

shewitt-au:
@paxcut 

It's not like you beat around the bush. You know I think my code is safer. And I've got no control over the situation.

paxcut:
> @paxcut

> It's not like you beat around the bush. You know I think my code is safer. And you've said you do too. What more would I say? You're immovable, I'm immovable, and I've got no control over the situation.

How am I beating around the bush? I have said all I needed to say and you claim to understand it all. You can say anything thats on your mind. Most of the things that affect us escape our control. When we think we have control we are only fooling ourselves.
I make lots of mistakes and I always make sure to dwell on them in order to learn. I have a very open mind and more patience than anybody I know. Do you think of yourself as immovable and is that a good thing?

shewitt-au:
> How am I beating around the bush? 

What I meant is that you're not beating around the bush. Sorry if I wasn't clear. I meant that you have made stance clear.

shewitt-au:
A subset of this PR has been PRed [here](https://github.com/WerWolv/PatternLanguage/pull/171#issuecomment-2955238177)

shewitt-au:
I'm not even going to bother commenting. I will delete my posts as not to make your PR look like a bar fight. I will not respond, in any way, to any more correspondence on this issue. Don't mistake this for me thinking that you shouldn't have attributed appropriately (as I see it).

PR #171, opened 2025-06-06
fix: fixes for various errors related to shared_from_this.

paxcut:
A while back there were some changes to the pattern language library that changed the way shared_pointers are created using shared_from_this(). Unfortunatelly the changes were not complete and various bugs were created among them 2234, json type not working, unable to export files, static arrays of bitfields,... The cause of the errors was that in class Pattern the member m_parent was left as a raw pointer and it needs to be handled by shared pointers. Also there were some cases in which share pointers were needed but unique pointers were used instead. Both cause crashes when shared_from_this is used on pointers that are not managed by shared_ptr. Another source of errors were infinite loops of clone and reference that caused stack overflow. The fixes include making m_parent a weak pointer, turning unique pointers into shared pointers and moving codefrom the copy constructors into clone to break the infinite loops.These changes are the bare minimum needed to bring the pattern language back to the full functionality that it had before shared_from_this was introduced or at least thats the hope.

### DIFF
--- a/lib/include/pl/api.hpp
+++ b/lib/include/pl/api.hpp
@@ -126,7 +126,7 @@ namespace pl::api {
     /**
      * @brief A function callback called when a custom built-in type is being instantiated
      */
-    using TypeCallback = std::function<std::unique_ptr<ptrn::Pattern>(core::Evaluator *, const std::vector<core::Token::Literal> &)>;
+    using TypeCallback = std::function<std::shared_ptr<ptrn::Pattern>(core::Evaluator *, const std::vector<core::Token::Literal> &)>;
 
     /**
      * @brief A type representing a function.
--- a/lib/include/pl/patterns/pattern.hpp
+++ b/lib/include/pl/patterns/pattern.hpp
@@ -116,7 +116,13 @@ namespace pl::ptrn {
         }
 
         virtual std::shared_ptr<Pattern> clone() const = 0;
-        std::shared_ptr<Pattern> reference() { return shared_from_this(); }
+        std::shared_ptr<Pattern> reference() {
+            auto weakPtr = weak_from_this();
+            if (weakPtr.expired()) {
+                core::err::E0001.throwError("Cannot call shared_from_this if this is not shared.");
+            }
+            return shared_from_this();
+        }
 
         [[nodiscard]] u64 getOffset() const { return this->m_offset; }
         [[nodiscard]] virtual u128 getOffsetForSorting() const { return this->getOffset() << 3; }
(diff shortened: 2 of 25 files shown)


shewitt-au:
Wow!

paxcut1

paxcut2

>showed no interest in collaborating

Go back and read the PR.

paxcut:
> My post to this PR was a polite statement of my feeling that my work was being used without credit. 

Ok, lets see.

> I have to admit to not being happy with the fact this code is based on my work without any attribution or acknowledgment. I spent a lot of time on it. Given it is a whittled down version of my work, a link to the source PR would seem reasonable and in the spirit of collaborative development.

The feeling that you had that your work was being used without credit was misplaced. As a result you posted statements that implied misconduct on my part.  You had no justification to allege misconduct from me. That is not a polite thing to do.If you really had good intentions, you would have first asked about the code's origins before making any allegations. 

> You relpy rudely and with sarcasam.

And here you are again attributing my remarks without indicating exactly how I was rude or where the sarcasm lies. Based on what I said in this post my response was appropriate and to the point. I need to defend myself against false allegations and I need to present evidence to back up my defense. This could have been avoided if you wouldn't have jumped the gun and started making false allegations right off the bat. Even after that you are escalating the conflict instead of admitting that your allegations were false and apologizing for it. There are several comments posted in your pr that are also aimed at making me reply negatively as a way to start conflict. I ignore them for what they are and will continue to do so but false allegations cross a line. I have explained all my actions to you and you have claimed to  understand and accept them. that should be the end of the contingency as further elevations cannot be justified.

shewitt-au:
OK. After this post I'm done with this.

By your own admission in the screenshots I attached, the code was based on mine with bits you deemed unnecessary removed. Any doubts about this can be quickly verified by anyone that cares to look by examining the histories of our respective PRs.

I put a lot of effort into it, and you know that, you were active in the comments in my PR.

You look code that I wrote, put it your own PR and don’t bother acknowledging the time and effort I put in.

When I point that out you argue.

I see no point in arguing with you on this. If it isn’t obvious already that your conduct in this matter is questionable, to say the least, than I can’t image what I can say to make you do otherwise.

paxcut:
Part of your problem is that you dont seem to read the post I made in order to understand why I say that the pr code is not based on your code. I did write code based on your and then i started from scratch again when I felt that you showed no interest in collaborating. I even quoted myself the exact sentence describing how I created the code.I will quote it again now. Please read it before saying any more false accusations that you have no right to make.

"Starting from master i added code to reference() that throws an error if the weak pointer is expired. that finds all the cases when shared pointers need to be created.
I have fixed all the cases related to the four examples above and all the code runs just as before except for maybe the json example where the reference use zeroes it out which occurs in your version as well. I need to investigate further on that."

Starting from master means that any previous modifications were deleted and that there was no code left that belonged to your pr. You can ignore me or the things I say. you cannot state false claims based on that though because you are falsely accusing me. You are not the victim. Read all the posts and try to understand what is in them instead of assuming you know what i going on. 
Now you are saying that my conduct is questionable but in fact it is your conduct that needs to be put in chaeck as any objective assessment would clearly find.

PR #178, opened 2025-07-10
Accurate column error info with tabs. Avoid copy of pattern string.

shewitt-au:
### Problem description
The lexer does not handle handle tabs correctly. In some cases this results in incorrect columns in errror messages. Currently `PatternLanguage::executeString`  does a rewrite of the pattern string like this:

```
code = wolv::util::replaceStrings(code, "\r\n", "\n");
code = wolv::util::replaceStrings(code, "\t", "    ");
```

This does not handle pattern files like this:

```
#pragma author WerWolv
#pragma description Microsoft PE Portable Executable (exe/dll)

#pragma MIME application/x-dosexec
#pragma MIME application/x-msdownload
#pragma MIME application/vnd.microsoft.portable-executable

import std.core;
import std.string;
import type.guid;
import type.time;
//3456789
;;**destruct DOSHeader {
```

Note:
"**" is a single tab `char` (hard coded to four spaces in ImHex). It is shown as two characters to illustrate the space it occupies.

Error:
```
E: error: Type destruct has not been declared yet.
E:   -->   in <Source Code>:13:7
E: 13 | ;;    destruct DOSHeader {
E:            ^^^^^^^^
```

Link to org fix for this issue:
https://github.com/WerWolv/PatternLanguage/commit/f29a7fda0e6c4651f1e1667fd0d82c791afc6c1a#diff-715e5fce329a964b5f0cae4fa2d45a4273ba388614701f95a33a0e7750b1e9f9

### Implementation description
The lexer has been modifed to handle tabs. Multiple newline encoding support has also been added. This fixes the wrong columns in errors as described above, and as a bonus removes the need to copy the pattern-code string in `PatternLanguage::executeString`, and instead use a `const &`

### Other stuff
In the lexer before these changes:
`m_longestLineLength`
The line length of the first line of a pattern file is one `char` short (compared to subsequent lines).  This is due to `m_lineBegin` pointing to the `\n` of the previous line, except for the first line, where it points to the first `char` of the line. This is special-cased in `Lexer::location()` but not handled in the longest-line-length code.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,15 +71,6 @@ jobs:
         cd build
         ctest --output-on-failure
 
-    - name: ✉️ Run ImHex-Patterns tests
-      if: ${{ env.DISPATCH_TOKEN != '' }}
-      uses: mvasigh/dispatch-action@main
-      with:
-        token: ${{ secrets.DISPATCH_TOKEN }}
-        repo: ImHex-Patterns
-        owner: WerWolv
-        event_type: run_tests
-
      # Generate report from all gcov .gcda files
     - name: 🧪 Generate coverage report
       run: |
@@ -91,3 +82,8 @@ jobs:
         fail_ci_if_error: true
         token: ${{ secrets.CODECOV_TOKEN }}
         file: coverage_report.xml
+
+  imhex_patterns_tests:
+    name: 🧪 ImHex Patterns Tests
+    needs: tests
+    uses: WerWolv/ImHex-Patterns/.github/workflows/tests.yml@master
--- a/generators/include/pl/formatters/formatter.hpp
+++ b/generators/include/pl/formatters/formatter.hpp
@@ -28,7 +28,7 @@ namespace pl::gen::fmt {
 
     class FormatterPatternVisitor : public pl::PatternVisitor {
     public:
-        virtual ~FormatterPatternVisitor() = default;
+        ~FormatterPatternVisitor() override = default;
 
         void enableMetaInformation(bool enable) { this->m_metaInformation = enable; }
         [[nodiscard]] bool isMetaInformationEnabled() const { return this->m_metaInformation; }
--- a/lib/include/pl/core/preprocessor.hpp
+++ b/lib/include/pl/core/preprocessor.hpp
@@ -113,6 +113,10 @@ namespace pl::core {
         }
 
         void appendToNamespaces(std::vector<Token> tokens);
+        void saveTokens(api::Source *source, const std::vector<Token> &tokens);
+        const std::map<std::string, std::vector<Token>> &getParsedImports() const {
+            return m_parsedImports;
+        }
 
     private:
         Preprocessor(const Preprocessor &);
@@ -137,6 +141,7 @@ namespace pl::core {
         void registerStatementHandler(const Token::Keyword &name, auto memberFunction);
         void reportError(const std::string &message, const std::string &description);
 
+        std::map<std::string, std::vector<Token>> m_parsedImports;
         std::unordered_map<std::string, api::PragmaHandler> m_pragmaHandlers;
         std::unordered_map<Token::Directive, api::DirectiveHandler> m_directiveHandlers;
         std::unordered_map<Token::Keyword, api::StatementHandler> m_statementHandlers;
--- a/lib/include/pl/core/token.hpp
+++ b/lib/include/pl/core/token.hpp
@@ -220,8 +220,10 @@ namespace pl::core {
             constexpr bool operator==(const Comment &) const = default;
         };
 
+        using LiteralVariantType = std::variant<char, bool, u128, i128, double, std::string, std::shared_ptr<ptrn::Pattern>>;
 
-        struct Literal : std::variant<char, bool, u128, i128, double, std::string, std::shared_ptr<ptrn::Pattern>> {
+
+        struct Literal : LiteralVariantType {
             using variant::variant;
 
             [[nodiscard]] std::shared_ptr<ptrn::Pattern> toPattern() const;
--- a/lib/include/pl/helpers/variant_type_index.hpp
+++ b/lib/include/pl/helpers/variant_type_index.hpp
@@ -1,0 +1,31 @@
+#pragma once
+
+#include <variant>
+#include <type_traits>
+#include <cstddef>
+
+namespace pl::hlp {
+
+template<typename T, typename Variant>
+struct VariantTypeIndexImpl;
+
+template<typename T, typename... Ts>
+struct VariantTypeIndexImpl<T, std::variant<Ts...>> {
+    static constexpr auto value = [] -> std::size_t {
+        static_assert(
+            (std::is_same_v<std::remove_cvref_t<T>, Ts> || ...),
+            "T not in std::variant");
+        constexpr bool matches[] = { std::is_same_v<std::remove_cvref_t<T>, Ts>... };
+
+        for (std::size_t i=0; i<sizeof...(Ts); ++i) {
+            if (matches[i]) { return i; }
+        }
+
+        return -1;
+    }();
+};
+
+template<class T, class Variant>
+inline constexpr std::size_t VariantTypeIndex = VariantTypeIndexImpl<T, Variant>::value;
+
+} // namespace pl::hlp
--- a/lib/include/pl/pattern_language.hpp
+++ b/lib/include/pl/pattern_language.hpp
@@ -93,7 +93,7 @@ namespace pl {
          * @param checkResult Whether to check the result of the execution
          * @return True if the execution was successful, false otherwise. Call PatternLanguage#getCompileErrors() AND PatternLanguage#getEvalError() to get the compilation or runtime errors if false is returned
          */
-        [[nodiscard]] bool executeString(const std::string &code, const std::string &source = api::Source::DefaultSource, const std::map<std::string, core::Token::Literal> &envVars = {}, const std::map<std::string, core::Token::Literal> &inVariables = {}, bool checkResult = true);
+        [[nodiscard]] bool executeString(const std::string& code, const std::string& source = api::Source::DefaultSource, const std::map<std::string, core::Token::Literal> &envVars = {}, const std::map<std::string, core::Token::Literal> &inVariables = {}, bool checkResult = true);
 
         /**
          * @brief Executes a pattern language file
--- a/lib/include/pl/pattern_visitor.hpp
+++ b/lib/include/pl/pattern_visitor.hpp
@@ -4,6 +4,7 @@ namespace pl {
 
     namespace ptrn {
 
+        class Pattern;
         class PatternArrayDynamic;
         class PatternArrayStatic;
         class PatternBitfield;
@@ -28,6 +29,7 @@ namespace pl {
 
     class PatternVisitor {
     public:
+        virtual ~PatternVisitor() = default;
         virtual void visit(ptrn::PatternArrayDynamic& pattern)  = 0;
         virtual void visit(ptrn::PatternArrayStatic& pattern)   = 0;
         virtual void visit(ptrn::PatternBitfield& pattern)      = 0;
--- a/lib/source/pl/core/error.cpp
+++ b/lib/source/pl/core/error.cpp
@@ -40,7 +40,7 @@ namespace pl::core::err::impl {
                     }
                 }
                 errorLine = errorLine.substr(start, end - start);
-                arrowPosition = column - start;
+                arrowPosition = column - start - 1;
             }
 
             result += fmt::format("{}{}\n", lineNumberPrefix, errorLine);
--- a/lib/source/pl/core/lexer.cpp
+++ b/lib/source/pl/core/lexer.cpp
@@ -169,7 +169,7 @@ namespace pl::core {
         m_cursor++; // Skip space
         auto location = this->location();
 
-        while (m_sourceCode[m_cursor] != '\n' && m_sourceCode[m_cursor] != '\0') {
+        while (m_sourceCode[m_cursor] != '\n' && m_sourceCode[m_cursor] != '\r' && m_sourceCode[m_cursor] != '\0') {
 
             auto character = parseCharacter();
             if (!character.has_value()) {
@@ -192,7 +192,7 @@ namespace pl::core {
 
         while (m_sourceCode[m_cursor] != '\"') {
             char c = peek(0);
-            if (c == '\n') {
+            if (c == '\n' || c == '\r') {
                 m_errorLength = 1;
                 error("Unexpected newline in string literal");
                 m_line++;
@@ -333,7 +333,7 @@ namespace pl::core {
         m_cursor += 2;
 
         std::string result;
-        while(m_sourceCode[m_cursor] != '\n' && m_sourceCode[m_cursor] != '\0') {
+        while(m_sourceCode[m_cursor] != '\n' && m_sourceCode[m_cursor] != '\r' && m_sourceCode[m_cursor] != '\0') {
             result += m_sourceCode[m_cursor];
             m_cursor++;
         }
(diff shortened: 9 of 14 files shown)


shewitt-au:
@paxcut 
> I can't find any evidence of the errors not handling spaces together in your example or any other file

Perhaps I should have been more specific. Here's how I get the issue:

- Open a file (I use a PE file)
- Open the pattern file that already has tab in it.

That last bit is important. I modified pe.hexpat to contains tabs.

paxcut:
> That last bit is important. I modified pe.hexpat to contains tabs outside of ImHex.

Why sis you modify it? pe.hexpat is known to have been created using tabs and it is stored with tabs. It was used as input example when the changes to eliminate tabs from the text editor were made. It is currently impossible to have an opened file in the text editor that contains tabs. Allowing tabs to be added to the test editor creates many more problems that the one you encountered so they are not allowed to be used and are always converted to spaces before the code is added to the text editor.

> I guess they're not necessary. Column numbers being off in certain edge cases is not the end of the world. But given it was quite easy to fix I gave it a go.

If column numbers were wrong under any circumstance then that would require changes to make sure they are correct. I reran the tests in all three versions of imhex using pe.hexpat and in all cases the column numbers were correct. When tabs are introduced in the text editor it is easy to detect them because using the left and right arrow keys skip the entire tab but with spaces you move one char at a time.If tabs were introduced the changes here alone would not have fixed all the problems that tabs create so the only possible way to fix all the issues tabs create is to ensure they are never added  and are always converted to spaces beforehand.

>  In my opinion handling the character level parsing belongs in the lexer. I believe it to be more robust.

That's not an opinion, that's a know fact about lexers. Lexers convert the character stream into token streams so they need to parse the character stream. Languages also have character sets that list all possible characters that can be used to create valid language statements. Tab characters are not part of the pattern language character set so adding support for them in the lexer makes no sense.

The changes to remove tabs were done months ago and all the tests I run in previous versions show the right column numbers in the errors using any known form to add code in the text editor and trying to create a tab character. you can import, paste , type or drop text into the editor and the tabs will be converted into spaces before the editor gets to display them.

That being said it is possible that some of the changes never made it to master but a few hours ago I updated master to include all the changes and fixes that I had in my local build so now they should be included without any doubt.

shewitt-au:
Ok, I'll take this a bit at a time. Sorry if that's a pain:

paxcut

> Why sis you modify it? pe.hexpat is known to have been created using tabs and it is stored with tabs. It was used as input example when the changes to eliminate tabs from the text editor were made.

I modified it (outside of ImHex) for two reasons. The first was for testing purposes. The second is that, and I mean no insult here, there are better text edits than ImHex's.

> Allowing tabs to be added to the test editor creates many more problems that the one you encountered 

I sympathise here. Problems are obvious a pain in the ass. And it's more than possible I'm underestimating them. That said a text edit should support tabs. There seems to be a gentlemen's agreement between the lexer and the editor: I don't parse tabs, so don't give them to me. This is [sematic coupling](https://programmingbrain.com/2019/01/semantic-coupling). If in future the pattern editor is upgraded to support tabs (and a programmer's editor should) now the lexer and the editor need to be changed. Also on this, I've read requests from users asking for featues like auto-reload so they can use an external editor. Closing the gap between the editor of choice and ImHex's should be made easier if possible. Finally on this, I've not actually changed the editor's behaviour. But, hopefully, the lexer will be more cooperative if this it is in future.

> That's not an opinion, that's a know fact about lexers. Lexers convert the character stream into token streams so they need to parse the character stream. Languages also have character sets that list all possible characters that can be used to create valid language statements. Tab characters are not part of the pattern language character set so adding support for them in the lexer makes no sense.

Respectfully, I think the lexer should support tabs. The lexer parses files and tabs live in some files.

shewitt-au:
@paxcut

>  I reran the tests in all three versions of imhex using pe.hexpat and in all cases the column numbers were correct

I'm not sure what's going on here. I've attached a video.


https://github.com/user-attachments/assets/10eede46-fea1-470e-90b0-08d1f99854db

shewitt-au:
@paxcut 

> I'm not sure I follow what this sentence means. Pressing the tab key on column 1,2,3 and 4 takes you to the exact same location in the text editor so you must be meaning something else. I don't get those error messages with the wrong columns and I can't insert tab characters (\t) in the text editor.

I loading a pattern that already contains tabs, not adding tabs in the ImHex editor. Take care not to let the build overwrite your pattern. I did explain this. I'll try to be more clear in future.

shewitt-au:
@paxcut 

> Now lets look at your error. You neglected to say that the imported file had to be changed so that it contained two valid characters, followed by a tab and followed by valid characters that will trigger an error when evaluating. Under those conditions I see a problem but not where you thing it is at. the column is erroneously set to 7 but notice that the gap in the error line is longer than the one in the test editor.

Given that I've put breakpoints in both versions of the lexer (mine and the status quo) this seems unlikely.

shewitt-au:
@paxcut 

> Under those conditions I see a problem but not where you thing it is at. the column is erroneously set to 7 but notice that the gap in the error line is longer than the one in the test editor.

```
#pragma author me
u32 a = 1+1;
;***aaa
```

Error
```
E:   -->   in <Source Code>:3:6
E: 3 | ;    aaa
E:          ^^^
```

*** = one tab char

Breakpoint:
 `hlp::CompileResult<std::vector<Token>> Lexer::lex(const api::Source *source)`

```
        m_longestLineLength = std::max(m_longestLineLength, m_cursor - m_lineBegin);
        addToken(makeToken(Separator::EndOfProgram, 0));

        return { m_tokens, collectErrors() }; // <--- Breakpoint here
    }
```
db1

> Under those conditions I see a problem but not where you thing it is

Looks like it's fro the lexer.

paxcut:
What I am getting at is that the lexer is not receiving tabs. If it did the column numbers would be smaller instead of bigger because a tab is only 1 character. The only way the lexer can receive tabs is by removing the incorretly use function that switches tabs with 4 spaces in executeString. that code needs to be replaced to use what the rest of ImHex uses to deal with tabs and deal with them the correct way. The function is called replaceTabsWithSpaces() and it is usually part of a preprocessText() function that does also changes line ends. That's how you need to fix the error.

shewitt-au:
paxcut 

> What I am getting at is that the lexer is not receiving tabs

It is.

```
hlp::CompileResult<std::vector<Token>> Lexer::lex(const api::Source *source) {
        this->m_sourceCode = source->content;
        this->m_source = source;
        this->m_cursor = 0;
        this->m_line = 1;
        this->m_lineBegin = 0;
        this->m_longestLineLength = 0;

        const size_t end = this->m_sourceCode.size();

        m_tokens.clear();

        while (this->m_cursor < end) {
            const char& c = this->m_sourceCode[this->m_cursor];

            // Debugging
            if (c=='\t')
                __debugbreak();
```

itis

paxcut:
No it isn't. The only way for the lexer to receive tabs is if you remove the lines in executeString() that convert the tabs to spaces. You can't change the code to make it do what you want to show as being the problem. 

At any rate the bug is fixed as I show in the video. 

> there is only one place to properly handle tabs. The lexer.

No it isn't and it is very easy to prove so. look at the code that fixes the error and you wont find any tab handling in the lexer. The tabs are properly handled, that means no 4 spaces for each tab. the handling is done before the lexer gets to see it so no tab handling for the lexer. QED.

You are obviously not going to implement the fix the proper way and are going to continue insisting your way is the only way. I then have no choice again but to create my own pr and fix the problem the way it needs to be fixed. Why is this the way to fix it? Analyze the error. It is created by the substitution of tabs with 4 spaces when it needs to choose how many spaces it needs based on its location. that's all there is to it. short and simple remove two lines and add one. replace the bad char substitution with the good one.Moving the char substitution to the lexer is not part of fixing the bug. It is there because the author put it there and he had his reasons to do so.If you think thats wrong create a pr for that but don't mix it with fixing the bug. 

https://github.com/user-attachments/assets/c920edea-cab1-40aa-a8e0-fcd55ec29e4e

shewitt-au:
paxcut
> You are obviously not going to implement the fix the proper way and are going to continue insisting your way is the only way. I then have no choice again but to create my own pr and fix the problem the way it needs to be fixed. 

I am no more quity because I think my way is better than you are for thinking yours is.

paxcut
> No it isn't. The only way for the lexer to receive tabs is if you remove the lines in executeString() that convert the tabs to spaces. You can't change the code to make it do what you want to show as being the problem.

I did't fake the test if that's what you mean by, "You can't change the code to make it do what you want to show as being the problem."

On the tab, if you mean lexing performed for the editor then yes. But as I mentioned ealier there are many paths to the lexer and not all go through `executeString`.

```c++
void ViewPatternEditor::loadPatternFile(const std::fs::path &path, prv::Provider *provider) {
        wolv::io::File file(path, wolv::io::File::Mode::Read);
        if (file.isValid()) {
            auto code = file.readString();
            // Debugging
            std::cout << code << std::endl;

            // TODO: lextwice: we lex twice here! 1st lex.
            this->evaluatePattern(code, provider); // TODO: This seems to be highlighting.
            m_textEditor.get(provider).SetText(code, true);
            m_sourceCode.get(provider) = code;

            // TODO: lextwice: we lex twice here! 2nd lex.
            TaskManager::createBackgroundTask("hex.builtin.task.parsing_pattern", [this, code, provider](auto&) { this->parsePattern(code, provider); });
        }
    }
```

`parsePattern` does no tab replacements. There may be other paths to the lexer I'm not aware of. Tabs do make it into the lexer and it's easy to verify.

paxcut:
> I am going to finish this PR. The last issue is the longest line length. After that I can't control what is done with it.

As I explained in the code comments the function that returns the longest line length should return exactly that and not the adjusted length to see 2 chars beyond the end. 

> I was referring to the fixes. Not the code base as a whole. I understand that you may not want to go against the intent of the author. But I find it hard to believe fixing the problem in the lexer violates the intent, although it may. I don't know.

What fixes? the one where I replace the calls to replaceString with a call to replaceTabsWithSpaces?  how is that not his code adjusted to fix the problem you found? He posted the original fix and placed the tab substitution in executeString and not in the lexer. so your code is going against his fixes. and yes it violates the intent as we discussed the issue which led to him posting the fix.

> Fixing the problem anywhere but the lexer, in my opinion, is just asking for troubles in the future

the difference is that we have experienced the problems that come with allowing tabs to enter imhex and we found a way to solve them and this is not an issue with the lexer. that's your focus because that happens to be the code you wrote. The reasons you give sound like good reasons bur not when weighed against the problems that tabs create in imhex. String preprocessing is what you are doing but in the wrong place. it needs to be done before the code enters the main evaluation engine that starts with the lexer. that's the only way to avoid the issues the original commit addresses. String copying is not even worth mentioning here. Errors are being reported accurately and without fuss. Bugs will happen and they will be fixed when found so that is always the truth. it is possible to introduce errors in the lexer parsing also. no code is immune to bugs. you cannot write code now to address future code issues. if you can predict the future maybe. It isnt less fragile. you are not taking the problems the original commit addresses into consideration in your evaluation and that makes your solution a lot more fragile than the current one where no tabs exist in Imhex and all code locations are reported correctly. the lexer only sees spaces.

> I am not being stubborn and refusing to see the "proper way". I've explained my reasons and they make sense.

i dont know if your are being stubborn or not. I know you are not accepting the solution that the author wrote and you think you know better than him. At least that's the impression I get when reading your posts. You may be right, I don't really know. I can only judge evidence that I know about that means make minimal changes that fix the errors. that leaves original code like it was before and fixes the new issues found. To evaluate your changes positively I need a leap of faith that I can't make. From what i understand, trouble will come that is best avoided by doing what we do.

WerWolv:
I didn't read though this whole arguing tbh but my stance is, and I discussed that with paxcut on Discord, that the pattern language library needs to be able to handle whatever's thrown at it. Even though ImHex specifically filters out these characters, they can still end up in the library if it's not called from within ImHex.
Up until that commit, all I did was to just get rid of carriage returns and tabs before feeding the data into the lexer which is obviously less than ideal. 
So the tab handling code I think would be great to add now at this point

paxcut:
yes, i still object to the code in this pr. I was never given a chance to voice my opinion on those changes so I'll do it now. There are two strategies to deal with this issue. you either fix every problem they create or you don't allow the problem to exist in the first place. The patching of the lexer is not enough to fix the problems that occur outside of the lexer, even if it is only for the pattern language. This 'fix' will have to be revisited again and again. the fix i propose is a one time thing and doesn't need to be looked at again.

paxcut:
The parser doesn't handle everything you send its way. When it detects an error it gives a message. that's how errors are handled, not by silently allowing the bad code to remain in the source code. With the changes here we will have no way of knowing if bad characters are making it past the filters so it is a bad idea to simply parse them. they should be treated like any other errors in the code because they are illegal characters.

paxcut:
That's not three, you are describing the first one. you think the lexer is the only one having problems but you can't be sure. Beside, you dont seem to understand why the code was added, it has nothing to do with the error this pr fixes. it arises because there are tests done that bypass al the filters and feed raw text straight to the evaluation. that is going to present problems if tests are done that use the bad characters that don't involve the lexer. 
Imhex is free of tab and \r problems thanks to the early elimination of them from the character stream and the pattern language should be using the same proven implementation to deal with the same problem

paxcut:
look, it is simple. you either deal with it early, or you deal with it at the end. thats two strategies, not three. early means as soon as the characters are read, at the end is when the errors occur. in this case the errors we know of are at the lexer so thats at the end. if errors occur somewhere else we'll have to fix them again and again. if you fix them early you never again have to deal with them. I have explained this over and over and it is simple and clear and can't be argued against. not only patching the parser wont fix errors occurring elsewhere but that hides the fact that those characters exist in the input files. making it harder to find and fix the new errors.

WerWolv:
I still disagree with this Ax, if we don't properly handle these characters in the Lexer or at least shortly before the Lexer, we will have to fix this same issue everywhere. We need to guarantee that any code that passes any pattern language code to the library needs to not contain any carriage returns or tabs which moves the responsibility out to the user who now needs to know about this and needs to handle it. If we just handle these characters properly ourselves, everybody using the library, including us in ImHex, will have it _just work_. And I am personally a big fan of libraries that _just work_

shewitt-au:
This is the last I will comment on this (extenuating circumstances aside), and here it is:

The lexer not handling tabs and various new line encodings in the lexer is obviously just bad design. I really can't be bothered trying to convince you of the obvious any more. It just seems pointless.

paxcut:
It is a bad idea because there may be other parts affected by those characters so the lexer should issue and error just like when it detects a bad identifier. There is plenty of character streams that the lexer cannot lex. it only accepts a subset of all possible char streams as valid code. The question is not to accept as many as possible. It doesn't matter how you want to use the pattern language library. if it contains code that uses those bad characters in some form that causes errors then accepting them as valid is a bad idea. how can you guarantee that will never happen?

shewitt-au:
paxcut
> There is plenty of character streams that the lexer cannot lex.

Naturally. But it would handle all chars in the character set.

paxcut:
You want to deal with errors as they happen? fine. If it ever falls on me to fix problems about tabs or other chars I know how to fix them.  I have already spent many hours (days?) dealing with bad chars in ImHex as a whole and I have 5 or 6 prs to prove it. I have learned from that experience but nobody wants to hear what that is so they'll have to learn on their own. Imhex is already dealing with tabs and chars properly. 

The lexer running in imhex will never use the code you are adding. it will only be used in the unit tests and probably no even there since now all the test are passing.Future test may fall because the characters were not removed early. That can be avoided and I am showing how. removing them early is the only way to guarantee no further problems will exist. thats indisputable unless you can predict the future.

 I choose that option because it is clearly better to not have to deal with problems than having to. you choose the other, why? i really don't know and you don't say. how can be better to deal with problems than not having to deal with them? that sounds to me like the very definition of bad design.
 
  From a purely practical point of view there is no doubt that early patching is superior. If you want to include what you have learned at the university or elsewhere that's not just some heuristic or cliche then by all mean, explain in detail. I am very reasonable when it comes to logical statements that are well made and will change my opinion instantly it you can show your pov has merit. I have proven that many times in many interactions in the past and recorded here in github. None of the arguments I have heard so far sound convincing. then again this is not my decision to make so my opinion is not that important. Agree to disagree and move on.

WerWolv:
I'm just gonna step in here and stop this whole discussion.
It's honestly extremely frustrating reading all of that.

I'm going to take care of this issue my way which mostly aligns with the things shewitt has proposed.

paxcut:
What's unfathomably about the verifiable existence of the prs mentioned in my hours of days of work on fixing problems on tabs and other chars on imhex before you ever posted your fist post on github? check the facts before you accuse.

PR #194, opened 2025-08-16
New lexer

shewitt-au:
### New lexer

I have been working on a new lexer. In no hurry. Just hacking away at it when time permits. And it's somewhere near ready. It uses an external header-only lexing library ([lexertl17](https://github.com/BenHanson/lexertl17)). And it's fast! Really fast! I'm sure there are problems, but I have not found any yet.

ImHex changes [here](https://github.com/WerWolv/ImHex/pull/2409)

I've tested it using GitHub's CI tests on the PL, but in ImHex I could not. The changes need approval (didn't used to).

### Implementation details

I've used an external header-only lexing library which uses DFA tables for the lexing. In debug builds these are built at runtime. In Release build I've set things up to build them at compile time. This is not strictly necessary, but I figured may as well. Faster is faster.

@WerWolv I'd appreciate if you could check out the pre-build steps I've added for these changes. They seem to work, but I used a Google-and-tweak approach to get it to work.

For me at least (an older computer) the speed increase is dramatic. I'd be interested on seeing how it performs on other systems.


WerWolv:
Hey, thanks for the PR. Next time though, before starting on a big change like that, especially when adding new dependencies, it would be great if you could join e.g our Discord server or otherwise have a discussion with me so we can lay out what actually should be done.

Currently I'm not really sure what kind of issues you're solving by rewriting the custom-made lexer. Is it really too slow? As far as I know, a sizable portion is spent in the Parser and basically everything else in the evaluator while the lexer is more or less instantaneous unless you're giving it gigantic files I guess. So what exactly is the use case? I'm open for improvements but I'm not sure just rewriting it is the right approach as there has been quite some thought been put into the whole pipeline by many people

shewitt-au:
I understand your reservations. I would suggest you give it a try. If it improves things as much for you as it does for me then your resonable reservations might be reduced. I've never used Discord. Am I too old for Discord? Perhaps I can still learn new tricks.

WerWolv:
Having small features or fixes merged is usually no issue but for bigger things, not having a discussion first will usually lead to you implementing things that aren't necessary needed or done in a way that doesn't align with other things. I promise you, talking to me about what you're planning to do will increase the chances of having your changes merged thousandfold :)
You can join our Discord here if you like, it's ultimately not too much different from IRC:  https://discord.gg/X63jZ36xBY

I wasn't trying to say your PR is not needed, I'm just wondering what _actual problem_ it's trying to address. Like if Lexing previously took 2ms and now it takes 1ms, that is a 50% improvement that basically nobody will ever notice. That's why I'm wondering why you chose to rewrite it completely in the first place

shewitt-au:
> I'm just wondering what actual problem

For me, in release build,  the analyzing data task is twice as fast with these changes. I can feel the difference.

shewitt-au:
@WerWolv 

I've been using this PR in all my local builds. Here's why:

Old lexer:

https://github.com/user-attachments/assets/7258217d-7ded-486c-888f-3d7a81b356b8

New lexer:

https://github.com/user-attachments/assets/78738497-a767-4bbd-a431-35591ec41d2f

When making it I dumped the output of both lexers for all patterns and compared them. I've been using it for months. The usability difference is anything but negligible. For me it’s dramatic. I does have an external dependency (header only).

I would be interested in hearing any thoughts on it. Perhaps there is something I could do to make it more acceptable? As things are I can only guess as to why something that, for me at least makes such a big difference, seems to have attracted so little interest.

PR #204, opened 2025-11-30
[PoC] Implement eager binding for nested template arguments

### LINKED ISSUE #144: Incorrect evaluation occurs when a local variable name collides with the template argument of an instantiated template parameter.
``` c++
struct array<auto count> {
    u32 value [count];
};


struct root<type> {
    u8 size; 
    type a;
};


u8 size@1;

using a1 = array<size>;
root<a1> rot @ 0;
```

The template parameter size in array<size> should refer to the template argument, not the local variable inside root struct. 

Screenshot 2024-12-24 at 8 08 56 PM

AkiSakurai:
This commit changes the evaluation of nested template arguments from lazy (pass-by-name) to eager (pass-by-value) to enforce correct lexical scoping.

Motivation:

Previously, in a nested template like `T<K<x>>`, the argument `x` was resolved lazily within the scope of `T` instead of the scope where it was declared. This pass-by-name behavior for nested types violated lexical scoping and caused unpredictable variable resolution.

Implementation:

The core of the fix was to separate a type's declaration from its application:

1. A new `ASTNodeTypeApplication` node was introduced to represent a specific instantiation of a type. It captures the fully resolved template arguments at the declaration site, effectively creating a closure over the surrounding scope.

2. The parser was updated to eagerly evaluate all arguments—including nested types—into this new `ASTNodeTypeApplication` node.

3. `ASTNodeTypeDecl` was simplified to only handle the static declaration of a type, leaving all instantiation-specific logic to `ASTNodeTypeApplication`. As a key benefit of this cleaner separation, forward-declared types can now be used in templates, enabling mutually recursive template definitions.

4. A new test suite (`test_pattern_template_parameters_scope.hpp`) was added to verify the corrected behavior with various nested and recursive template scenarios that previously failed.

fix #144

### DIFF
--- a/cli/source/subcommands/docs.cpp
+++ b/cli/source/subcommands/docs.cpp
@@ -35,7 +35,7 @@ namespace pl::cli::sub {
         }
 
         std::string getTypeName(const core::ast::ASTNode *type) {
-            if(auto typeApp = dynamic_cast<const core::ast::ASTNodeTypeApplication*>(type); typeApp != nullptr) {
+            if (auto typeApp = dynamic_cast<const core::ast::ASTNodeTypeApplication*>(type); typeApp != nullptr) {
                 return fmt::format("{}{}", getTypeEndian(typeApp), getTypeName(typeApp->getType().get()));
             } else if (auto typeDecl = dynamic_cast<const core::ast::ASTNodeTypeDecl*>(type); typeDecl != nullptr) {
                 return typeDecl->getName();
--- a/lib/include/pl/core/parser_manager.hpp
+++ b/lib/include/pl/core/parser_manager.hpp
@@ -36,11 +36,11 @@ namespace pl::core {
 
         void reset() {
             this->m_onceIncluded.clear();
-            for(const auto &[_, types] : this->m_parsedTypes) {
-                for(const auto &[_, type] : types) {
-                    if(type != nullptr && type->isValid()) {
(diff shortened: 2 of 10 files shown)


WerWolv:
Hey! Wow thank you so much for that work! That has been a very long standing issue.

If you haven't already, could you run your changes on the patterns in the WerWolv/ImHex-Patterns repo? The unit tests there should provide good coverage to make sure nothing else broke and will show if there's any changes that will need to be done due to the changed lexical scope now.

WerWolv:
> It seems that the CI for ImHex Patterns Tests does not pull the correct branch.

That should be fixed now. The CI now passes along the current git hash and the git repo to the ImHex-Pattern repo unit tests CI and uses that repo instead of being hardcoded to master of this repo.
If you can rebase your PR, it should then properly run on the right repo

WerWolv:
Seems to fully pass now besides the Xilinx Bootgen Pattern you mentioned earlier. I will update that one manually once your PR is merged

WerWolv:
Yeah you're right, this is going to break patterns but I'm more than willing to do that because having to use parent there in the first place never made sense.
I will definitely update the docs and add a note about it to the next release notes. I think that should be alright. If we get support requests about it, it will be easy enough to explain

I'd expect not too many people to have used it before since it only worked for really basic cases. As you see from the unit tests now, there's only a single pattern in the ImHex-Patterns repo that was affected by this

PR #205, opened 2025-12-06
Update pattern_language_git_hash to use github.ref instead of github.sha in tests workflow

AkiSakurai:
github.sha does not work for pull requests.


WerWolv:
The issue seems to be that pattern_language_git_repo still gets populated with my repo url instead of yours

PR #209, opened 2025-12-10
evaulator: fix crash on identity using declarations

AkiSakurai:
This commit resolves a crash that occurred when processing using declarations that aliased a type to its template parameter, such as using T<K> = K. This was caused by a type not being resolved in getTypeDefinition. 

The fix ensures that type template arguments are passed down in getTypeDefinition, and extracts the actual type definition.


WerWolv:
Thanks a lot!

PR #211, opened 2025-12-14
evaluator: Optimize rvalue access

AkiSakurai:
Accessing elements of an rvalue array previously required creating a temporary shallow copy of the array or struct entries, which could be unnecessarily slow.

This commit eliminates the unnecessary copy by iterating over the elements directly.


WerWolv:
Thank you!

PR #223, opened 2026-03-09
fix: json format unit test.

paxcut:
Recent changes to json string format output produces a result different from the one saved in the data read to check the result. The expected string is changed to match the resulting formatted string.


paxcut:
already fixed

PR #244, opened 2026-08-25
fix: Make $[], sizeof($) and addressof($) respect the current section + added tests for it

ddmitv:
Before this fix, $[*address*] returned a byte from the main section, regardless of the current active section it's being executed in. I'm pretty sure this is a bug because `std::mem::read_unsigned` (and other functions in `std::mem`) actually respects the current section (see [here](https://github.com/WerWolv/PatternLanguage/blob/03dd3b0101b647a9c7aa658c709eb87cef5e9f75/lib/source/pl/lib/std/mem.cpp#L139)). `sizeof($)` and `addressof($)` were also fixed. They now work exactly the same way as [`std::mem::size`](https://github.com/WerWolv/PatternLanguage/blob/03dd3b0101b647a9c7aa658c709eb87cef5e9f75/lib/source/pl/lib/std/mem.cpp#L97-L103) and [`std::mem::base_address`](https://github.com/WerWolv/PatternLanguage/blob/03dd3b0101b647a9c7aa658c709eb87cef5e9f75/lib/source/pl/lib/std/mem.cpp#L86-L95).

Additionally, I added tests for the `$` operator since it didn't have its own tests before and I wasn't sure where to put tests for `$[]`, `sizeof($)` and `addressof($)`. Small note: `#pragma base_address` doesn't exist here and I'm not sure how to change base address without it, so `addressof($)` is not actually being tested with a modified base address.

### DIFF
--- a/lib/source/pl/core/ast/ast_node_rvalue.cpp
+++ b/lib/source/pl/core/ast/ast_node_rvalue.cpp
@@ -71,9 +71,10 @@ namespace pl::core::ast {
                         auto offsetLiteral = dynamic_cast<ASTNodeLiteral*>(offsetNode.get());
                         if (offsetLiteral != nullptr) {
                             auto offset = u64(offsetLiteral->getValue().toUnsigned());
+                            const u64 currentSection = evaluator->getUserSectionId();
 
                             u8 byte = 0x00;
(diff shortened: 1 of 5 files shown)


WerWolv:
Thank you!

PR #246, opened 2026-09-01
feat: Add a pluggable text codec for string patterns

scottanderson:
Add `StringEncodeDecode`: an interface with `decode()`, `encode()`, and `encodeLossy()`. The host application sets one on the Evaluator. With none set, a string pattern keeps its raw-byte behavior.

- `decode()` and `encode()` return `std::optional`. A `nullopt` result marks a byte sequence, or a character, the named encoding cannot represent.
- `encodeLossy()` never fails. It substitutes a replacement character for anything the encoding cannot represent.
- `PatternString` routes reads and writes through the codec. `getValue()` and `getBytesOf()` throw `core::err::E0004` on a `nullopt` result. A script can catch this error with try/catch.
- `setValueLossy()` writes through `encodeLossy()`, and clears the cached bytes and the cached display string; `getBytesOf()` otherwise reflects a write only after the next pattern run.
- `setValue()` clears the cached bytes the same way, on every pattern type, not just `PatternString`.
- `formatDisplayValue()` decodes through `decode()`. It returns the raw decoded text. The host application escapes the text for display.
- `getBytesOf()` caps the encoded result to the pattern's own size. 
- `getEncodingName()` reads the string's own `encoding` attribute.
- `PatternWideString` keeps its fixed UTF-16 behavior. The codec does not apply to it.
- `PatternLanguage::clearFormatCaches()` clears every placed pattern's cached display value, across every section.

### DIFF
--- a/lib/include/pl/core/string_encode_decode.hpp
+++ b/lib/include/pl/core/string_encode_decode.hpp
@@ -10,6 +10,20 @@
 
 namespace pl::core {
 
+    /// Why decode() stopped before the end of `bytes`.
+    enum class DecodeStop {
+        EndOfInput,      ///< Consumed every byte given
+        CodepointLimit,  ///< Reached maxCodepoints; more input may remain
+        MalformedBytes,  ///< The byte sequence at bytesConsumed is not valid under this encoding
+    };
+
+    struct DecodeResult {
+        std::string text;                              ///< Whole code points only, never a mid-sequence fragment
+        size_t bytesConsumed = 0;                       ///< How much of the input `text` accounts for
+        size_t codepointCount = 0;                      ///< Code points in `text`
+        DecodeStop stopReason = DecodeStop::EndOfInput;
+    };
+
     /**
      * @brief A pluggable text codec for string patterns
      * @note The host application sets this on the Evaluator. With none set, string
@@ -20,14 +34,17 @@ namespace pl::core {
         virtual ~StringEncodeDecode() = default;
 
         /**
-         * @brief Decodes `bytes` under `encoding`
+         * @brief Decodes `bytes` under `encoding`, one code point at a time
          * @param bytes Bytes to decode
          * @param encoding Encoding to decode with; empty when the pattern names no
          * encoding, in which case the codec picks its own default
-         * @return The decoded text, or std::nullopt when `bytes` is not valid under
-         * `encoding`
+         * @param maxCodepoints Stop after decoding this many code points. std::nullopt
+         * decodes as much of `bytes` as is valid, with no cap.
+         * @return Code points decoded up to the first malformed byte, the end of
+         * `bytes`, or maxCodepoints - whichever comes first. `stopReason` says which.
          */
-        [[nodiscard]] virtual std::optional<std::string> decode(std::span<const u8> bytes, std::string_view encoding) const = 0;
+        [[nodiscard]] virtual DecodeResult decode(std::span<const u8> bytes, std::string_view encoding,
+            std::optional<size_t> maxCodepoints = std::nullopt) const = 0;
 
         /**
          * @brief Encodes `text` under `encoding`
--- a/lib/include/pl/core/string_encode_decode.hpp
+++ b/lib/include/pl/core/string_encode_decode.hpp
@@ -1,0 +1,53 @@
+#pragma once
+
+#include <optional>
+#include <span>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include <pl/helpers/types.hpp>
+
+namespace pl::core {
+
+    /**
+     * @brief A pluggable text codec for string patterns
+     * @note The host application sets this on the Evaluator. With none set, string
+     * patterns keep their old raw-byte behavior.
+     */
+    class StringEncodeDecode {
+    public:
+        virtual ~StringEncodeDecode() = default;
+
+        /**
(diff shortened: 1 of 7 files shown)


lib/include/pl/core/string_encode_decode.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <span>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include <pl/helpers/types.hpp>
+
+namespace pl::core {
+
+    // A pluggable text codec for string patterns. The host application sets this on the
```
WerWolv: All comments that decorate classes or functions should use doxygen style

lib/include/pl/patterns/pattern_string.hpp
```diff
@@ -67,23 +123,39 @@ namespace pl::ptrn {
         }
 
         std::string formatDisplayValue() override {
-            auto size = std::min<size_t>(this->getSize(), 0x7F);
+            auto *evaluator = this->getEvaluator();
+            const auto fullSize = this->getSize();
+            auto size = std::min<size_t>(fullSize, 0x7F);
```
WerWolv: Is this truncating still okay to do now? We should probably try to not cut codepoints in half anymore

lib/include/pl/patterns/pattern_string.hpp
```diff
@@ -32,23 +35,76 @@ namespace pl::ptrn {
 
         }
 
+        // Empty when this pattern names no encoding. The codec then falls back to its own
+        // default.
+        [[nodiscard]] std::string getEncodingName() const {
+            if (const auto &arguments = this->getAttributeArguments("encoding"); !arguments.empty())
+                return arguments[0].toString(true);
+            return "";
```
WerWolv: I think we should store the default encoding in the evaluator and return that here if no [[encoding]] attribute is present. That way we can handle `#pragma encoding` in the runtime too and give applications the opportunity to set a default encoding

lib/include/pl/patterns/pattern_string.hpp
```diff
@@ -67,23 +123,39 @@ namespace pl::ptrn {
...
+                buffer.erase(pos + 1);
 
-            buffer.erase(pos + 1);
+                return fmt::format("\"{0}\" {1}", hlp::encodeByteString({ buffer.begin(), buffer.end() }), truncatedSuffix);
+            }
 
-            auto displayString = hlp::encodeByteString({ buffer.begin(), buffer.end() });
+            auto decoded = codec->decode({ reinterpret_cast<const u8*>(buffer.data()), buffer.size() }, this->getEncodingName());
+            if (!decoded.has_value())
+                throw std::runtime_error("Invalid");
```
WerWolv: Don't throw a runtime error here, use the regular error reporting functions

WerWolv (changes requested):

lib/include/pl/patterns/pattern_string.hpp
scottanderson: I attempted to improve this in the next revision

lib/include/pl/core/string_encode_decode.hpp
scottanderson: Done

lib/include/pl/patterns/pattern_string.hpp
scottanderson: Done

lib/include/pl/patterns/pattern_string.hpp
scottanderson: Fixed

lib/include/pl/patterns/pattern_string.hpp
```diff
@@ -67,23 +129,72 @@ namespace pl::ptrn {
...
+            }
+
+            const auto encoding = this->getEncodingName();
+
+            // The read above can itself end mid-codepoint. decode() fails on the whole
+            // buffer in that case, not just its incomplete tail, so back off a few
+            // bytes and retry rather than reporting a merely-truncated read as invalid.
+            std::optional<std::string> decoded;
+            for (size_t backoff = 0; backoff <= std::min<size_t>(4, buffer.size()); ++backoff) {
+                decoded = codec->decode({ reinterpret_cast<const u8*>(buffer.data()), buffer.size() - backoff }, encoding);
```
WerWolv: This feels _really_ expensive to do over and over. Maybe just passing it at most 8 bytes at the end would be enough in order to try and decode it and find the real end instead of the whole string? Alternatively, maybe adding a function to the codec to get the number of bytes that can successfully be encoded would make more sense. Then the encoding has to be done at most twice instead of 5 + 1 times

lib/source/pl/lib/std/pragmas.cpp
```diff
@@ -38,6 +38,10 @@ namespace pl::lib::libstd {
                 return false;
         });
 
+        runtime.addPragma("encoding", [](pl::PatternLanguage &runtime, const std::string &value) {
```
WerWolv, line 41: This is good. Just as a note for later, we will need to override this pragma again in ImHex so we can also adjust ImHex's encoding with that pragma

scottanderson: I had it working that way, but it was extremely buggy so I redesigned it to use an observer pattern instead. The latest commit in this PR has the PL changes for that.

WerWolv (approved):

PR #247, opened 2026-09-01
feat: Publish a per-run snapshot of string pattern encodings

scottanderson:
Add `StringEncodingRegion`: the address, size, and encoding of one string pattern. PatternLanguage builds a sorted snapshot of these after every successful run. It publishes the snapshot through an atomic `shared_ptr`.

`getStringEncodingRegions()` returns the full snapshot. `findStringEncoding()` looks up one exact address and size. Both functions are safe to call from any thread.

Depends on #246.


WerWolv:
This currently contains all the changes from #246 as well

PR #253, opened 2026-09-08
Slight updates to PL (before visitors)

Nemoumbra:
This adds more tests for various identifier clashes. Some of them are intentionally disabled, because the current validator can't catch them. The comments will be removed later.

Also some stringify-ish funcs for PL objects and a small change for the Result API (pass by value and move). We can later change the move ctor and move= too.


tests/include/test_patterns/test_pattern_error_semantics.hpp
```diff
@@ -123,4 +123,195 @@ namespace pl::test {
...
+    class TestPatternMultiVarRedeclarationFail: public TestPatternFailingSemantic {
+    public:
+        TestPatternMultiVarRedeclarationFail(core::Evaluator *evaluator)
+        : TestPatternFailingSemantic(evaluator, "MultiVarRedeclarationFail", R"(
+                struct S {
+                    u8 x, x;
+                };
+            )") { }
+    };
+//    class TestPatternMultiVarMixRedeclarationFail: public TestPatternFailingSemantic {
```
WerWolv, line 169: Why are these commented out?

WerWolv (changes requested):

tests/include/test_patterns/test_pattern_error_semantics.hpp
Nemoumbra: Would you like those to be uncommented straight away, even though they aren't meant to run as of this PR? My intention is to enable these tests after I submit the 2 validators I wrote to tweak and strengthen the logic.

WerWolv (approved):

PR #254, opened 2026-09-08
AST visitors

Nemoumbra:
This adds the a visitor framework for traversing the AST.
It consists of an abstract class (iface) `ASTVisitor` which implements `visit` for all the available AST node types.... +- the actual ASTNode... maybe we don't need this fallback... idk......... Also a pure virtual method `accept(ASTVisitor&)` for AST nodes. The visitor pattern, nothing fancy.
I also added getters for nodes which didn't expose some valuable fields.

One visitor example is the iface for writing AST validators, called `ASTValidator`. It implements all methods as empty, so you don't need to do that for all the children when you don't care about some nodes.
The AST validators are combined in a pipeline, which is meant to replace the current validator.

This branch is created from the previous one, so it has the commits from the last PR. I just couldn't wait to share this code.
I don't want to commit more until we sort out the includes, namespaces or filesystem locations for my additions. Then I'll follow with the 2 validators and also a CLI command to generate GV files (this is an example for non-validating visitor).


lib/include/pl/core/ast/visitors/ast_validator.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+        Result validate(const std::vector<std::shared_ptr<ast::ASTNode>> &ast);
+
+    protected:
+        virtual void reset() = 0;
+
+        // Helpers
+        template <ast::vis::Acceptable T>
+        void visitTypeErased(const T& node) {
+            auto node_ptr = node.get();
```
WerWolv: camelCase

lib/include/pl/core/ast/visitors/ast_validator.hpp
```diff
@@ -0,0 +1,118 @@
...
+            if (m_validatedNodes.contains(node_ptr)) {
```
WerWolv: Is there ever a case when we'd like to repeat visiting nodes? I guess if yes, we'd need a different way of detecting loops in the tree

lib/include/pl/core/validator.hpp
```diff
@@ -43,4 +45,17 @@ namespace pl::core {
         std::list<std::unordered_set<std::string>> m_identifiers;
     };
 
+    class ValidatorPipeline {
+        using Result = hlp::CompileResult<bool>; // not sure where to centralize the declaration
```
WerWolv: Here is fine I think. Though it should probably be public or at least protected

lib/include/pl/pattern_language.hpp
```diff
@@ -61,11 +62,12 @@ namespace pl {
...
         struct Internals {
-            std::unique_ptr<core::Preprocessor> preprocessor;
-            std::unique_ptr<core::Lexer>        lexer;
-            std::unique_ptr<core::Parser>       parser;
-            std::unique_ptr<core::Validator>    validator;
-            std::unique_ptr<core::Evaluator>    evaluator;
+            std::unique_ptr<core::Preprocessor>      preprocessor;
+            std::unique_ptr<core::Lexer>             lexer;
+            std::unique_ptr<core::Parser>            parser;
+            std::unique_ptr<core::Validator>         validator;
```
WerWolv, line 68: camelCase again.

Do we still need the old Validator after this?

lib/source/pl/pattern_language.cpp
```diff
@@ -34,13 +34,16 @@ namespace pl {
...
-            .evaluator      = std::make_unique<core::Evaluator>()
+            .preprocessor       = std::make_unique<core::Preprocessor>(),
+            .lexer              = std::make_unique<core::Lexer>(),
+            .parser             = std::make_unique<core::Parser>(),
+            .validator          = std::make_unique<core::Validator>(),
+            .validator_pipeline = std::make_unique<core::ValidatorPipeline>(),
+            .evaluator          = std::make_unique<core::Evaluator>()
         };
 
+        configureValidatorPipeline();
```
WerWolv: I think doing this in the constructor of the ValidatorPipeline would make sense?

lib/source/pl/pattern_language.cpp
```diff
@@ -185,7 +188,16 @@ namespace pl {
             this->m_compileErrors.insert(m_compileErrors.end(), validatorErrors.begin(), validatorErrors.end());
             validatorErrors.clear();
         }
+        // TODO: delete the block above when ready
+
+        auto [_, validatorPipelineErrors] = this->m_internals.validator_pipeline->validate(ast.value());
 
+        if (!validatorPipelineErrors.empty()) {
+            auto begin = std::move_iterator(validatorPipelineErrors.begin());
```
WerWolv: I think you could use `std::ranges::move(validatorPipelineErrors, std::back_inserter(m_compileErrors));`for this

WerWolv (changes requested):

lib/include/pl/pattern_language.hpp
Nemoumbra: Once I PR/commit to this branch the replacement, no. The replacement is ready, btw.

lib/source/pl/pattern_language.cpp
Nemoumbra: Well, I thought of it as a container. And containers don't typically populate themselves.

lib/include/pl/core/ast/visitors/ast_validator.hpp
Nemoumbra: I couldn't come up with any examples. I'd say it's very unlikely.

lib/source/pl/pattern_language.cpp
WerWolv: I mean more that you can initialize the pipeline directly like this: 
```cpp
std::make_unique<ValidatorPipeline>({
    ValidatorA(),
    ValidatorB()
});
```

WerWolv (approved):

PR #255, opened 2026-09-10
fix: prevent forEachEntrySorted crash when sorted entries are stale

AkiSakurai:
setFields() fills m_sortedFields, but setEntries() overwrites m_fields without touching m_sortedFields. If both get called on the same pattern, sorted goes stale and forEachEntryImpl reads past the end of the sorted vector.

Happens with template bitfields:
  1. ASTNodeBitfield::createPatterns() calls setFields() -> both m_fields and m_sortedFields get populated
  2. ASTNodeTypeDecl::createPatterns() calls setEntries(scope) to add template parameter patterns -> m_fields grows but m_sortedFields is not updated
  3. forEachEntryImpl loops i < getEntryCount() using the stale sorted vector -> heap-buffer-overflow

Fix: setEntries() clears the sorted cache so getSortedEntries() falls back to getEntries(). Also hardened PatternBitfield:
- setFields() clears m_sortedFields before populating (prevents duplicates if called more than once)
- forEachEntryImpl() clamps end to patterns.size(), like the other pattern types already do

Applied to PatternBitfield, PatternBitfieldArray, PatternStruct, and PatternUnion.

I have a rather large pattern that crashes without ASan. The sample below only crashes with ASan since the OOB read lands in valid heap memory.

Sample pattern (ASan required):

    bitfield TemplateBitfield<auto N> {
        unsigned a : 2;
        unsigned b : 3;
    };
    TemplateBitfield<4> bf @ 0;

Test case added: SetEntriesSortedConsistency

### DIFF
--- a/lib/include/pl/patterns/pattern_bitfield.hpp
+++ b/lib/include/pl/patterns/pattern_bitfield.hpp
@@ -383,6 +383,8 @@ namespace pl::ptrn {
         }
 
         [[nodiscard]] std::vector<std::shared_ptr<Pattern>> getSortedEntries() override {
+            if (this->m_sortedEntries.empty())
+                return this->getEntries();
             return this->m_sortedEntries;
         }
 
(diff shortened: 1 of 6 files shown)


WerWolv:
Thank you!

PR #256, opened 2026-09-10
feat: support using decl types for in/out

neptuwunium:
Brought up here: https://github.com/WerWolv/ImHex/pull/2890#discussion_r3971959254

But `using` decls don't work at all even for builtin types. This fixes that.

### DIFF
--- a/lib/source/pl/core/parser.cpp
+++ b/lib/source/pl/core/parser.cpp
@@ -2429,22 +2429,50 @@ namespace pl::core {
 
         if (inVariable || outVariable) {
             bool invalidType = false;
-            if (const auto builtinType = dynamic_cast<ast::ASTNodeBuiltinType*>(type->getType().get()); builtinType != nullptr) {
-                if (const auto valueType = builtinType->getType(); !Token::isInteger(valueType)
-                                                                   && !Token::isFloatingPoint(valueType)
-                                                                   && valueType != Token::ValueType::Boolean
-                                                                   && valueType != Token::ValueType::Character
-                                                                   && valueType != Token::ValueType::String)
+            auto visitingType = type;
+            auto declNestLimit = 128; // some high value so infinite recursion breaks
+            while (!invalidType && visitingType && declNestLimit-- > 0) {
+                auto checkType = visitingType->getType();
+
+                if (checkType == nullptr) {
                     invalidType = true;
-            } else if (const auto typeDecl = dynamic_cast<ast::ASTNodeTypeDecl*>(type->getType().get()); typeDecl != nullptr) {
-                if (const auto enumDecl = dynamic_cast<ast::ASTNodeEnum*>(typeDecl->getType().get()); enumDecl == nullptr) {
-                    invalidType = true;
+                    break;
                 }
-            } else {
+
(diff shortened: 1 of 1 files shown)


lib/source/pl/core/parser.cpp
```diff
@@ -2429,22 +2429,50 @@ namespace pl::core {
...
-                if (const auto valueType = builtinType->getType(); !Token::isInteger(valueType)
-                                                                   && !Token::isFloatingPoint(valueType)
-                                                                   && valueType != Token::ValueType::Boolean
-                                                                   && valueType != Token::ValueType::Character
-                                                                   && valueType != Token::ValueType::String)
-                    invalidType = true;
-            } else if (const auto typeDecl = dynamic_cast<ast::ASTNodeTypeDecl*>(type->getType().get()); typeDecl != nullptr) {
-                if (const auto enumDecl = dynamic_cast<ast::ASTNodeEnum*>(typeDecl->getType().get()); enumDecl == nullptr) {
+            auto visitingType = type;
+            auto declNestLimit = 128; // some high value so infinite recursion breaks
```
WerWolv: kek

This made me discover that this code is in fact valid and triggers the evaluation depth limit. I didn't think we could have recursive type definitions like that
```cpp
using A;

using B = A;
using A = B;

A a @ 0x00;
```

lib/source/pl/core/parser.cpp
```diff
@@ -2429,22 +2429,50 @@ namespace pl::core {
...
+                    break;
                 }
-            } else {
+
+                if (const auto typeDecl = std::dynamic_pointer_cast<ast::ASTNodeTypeDecl>(checkType); typeDecl != nullptr) {
+                    checkType = typeDecl->getType();
+                }
+
+                if (const auto usingDecl = std::dynamic_pointer_cast<ast::ASTNodeTypeApplication>(checkType); usingDecl != nullptr) {
+                    [[unlikely]] if (usingDecl == visitingType) {
```
WerWolv: [[unlikely]] goes after the if head. So between the ) and the {

lib/source/pl/core/parser.cpp
```diff
@@ -2429,22 +2429,50 @@ namespace pl::core {
...
+                        // realistically should never happen but you never know
+                        invalidType = true;
```
WerWolv, line 2449: And this can apparently happen, see comment above

lib/source/pl/core/parser.cpp
WerWolv: Give this some explicit type like `i32`. `auto` will resolve to `int` here but I think it would be better to be explicit

WerWolv (changes requested):

lib/source/pl/core/parser.cpp
neptuwunium: that's scary

lib/source/pl/core/parser.cpp
neptuwunium: oops, it's been a while since i had to use that attribute

WerWolv (approved):

PR #257, opened 2026-09-11
New validators

Nemoumbra:
Follow up to the 2nd PR: 2 validators for the pipeline. The old one has been removed.
Also enables the tests I added in the 1st PR... except for the duplicate enum test.
We already discussed this issue and how to resolve it, so it's not in the scope of this PR.
Let's see if it holds water.


lib/source/pl/core/ast/visitors/control_flow_validator.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <pl/core/ast/visitors/control_flow_validator.hpp>
+
+#include <pl/core/ast/ast_node_function_definition.hpp>
+#include <pl/core/ast/ast_node_control_flow_statement.hpp>
+#include <pl/core/ast/ast_node_try_catch_statement.hpp>
+#include <pl/core/ast/ast_node_while_statement.hpp>
+#include <pl/core/ast/ast_node_match_statement.hpp>
+#include <pl/core/ast/ast_node_conditional_statement.hpp>
+
+namespace pl::core {
```
WerWolv, line 11: From what I can tell, this only handles control flow statements in functions but not inside types, correct?

lib/source/pl/core/ast/visitors/scope_validator.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+    void ScopeValidator::visit(ast::ASTNodeBitfieldArrayVariableDecl &node) {
+        const auto& field_name = node.getName();
+        if (!newIdentifier(field_name)) {
+            error("Redefinition of identifier '{}'", field_name);
+        }
+    }
+
+    void ScopeValidator::visit(ast::ASTNodeEnum &node) {
+        // TODO: enums are broken
```
WerWolv, line 158: Broken how? This definitely needs to be fixed before merge

WerWolv:
Code wise this looks great!
We'll need to fix the issues in the current patterns though before merging

lib/source/pl/core/ast/visitors/control_flow_validator.cpp
Nemoumbra: Yes.

lib/source/pl/core/ast/visitors/scope_validator.cpp
Nemoumbra: Broken "duplicate enum entries should be handled in the parser as we discussed in Discord".

